PR #67, opened 2022-02-21
Do not remove downloads/.gitignore

mhdirkse:


cypress/downloads/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
celmoussaoui, line 1: Isnt it supposed to be `.*` instead of `*`

mhdirkse: I tested it by putting some arbitrary file in the downloads directory. After updating `.gitignore` as done here, the file was no longer shown with `git status`. Do you want me to change to `.*` and test that?

celmoussaoui: Yes, can you test it?

mhdirkse: `.*` does not do it. Only `*`.

celmoussaoui: Okay

Issue #73, opened 2022-02-22
You cannot disable the report generator


mhdirkse:
You cannot disable the report generator as can be explained with the following figure:

![ladybugSettingsDialog](https://user-images.githubusercontent.com/6784732/155141351-ecd75772-d3e7-4d43-af3b-14b4a195c884.jpg)

Steps to reproduce:
1. Select "Disabled" in the drop-down box.
2. Press the "save" button (number 2).
3. The settings dialog disappears. Reopen it.
4. You still see your choice. The report generator is disabled.
5. Press the synchronize button (number 1).
6. The report generator is now enabled, which is not OK. This means that the action of disabling was not communicated with the server.

I also tried the link in the example application (runs on port 80) that normally creates a report. It still does so with the report generator disabled.

celmoussaoui:
Factory resetting the report generator means it will be set to a predefined value, not the server value which, in this case, is `Enabled`. 

Communicating with the server is in fact done when saving. When making changes again and using the 'regular' reset (not the factory reset) button, it will get the value given by the server, hence it is saved.

PR #74, opened 2022-02-22
Test disabling and enabling report generator

mhdirkse:


mhdirkse:
Issue #73 has to be fixed before this PR can be finished.

PR #77, opened 2022-02-22
Also run on Edge

mhdirkse:


mhdirkse:
At this point, the action that runs the tests with Firefox fails to save videos. I think this is caused by Cypress bug https://github.com/cypress-io/cypress/issues/18415. I propose we accept for now that we do not have videos for Firefox. The tests with Firefox do succeed.

celmoussaoui (approved):

PR #96, opened 2022-03-14
If there are open reports, one of them is selected

mhdirkse:


mhdirkse:
I will add to this test that the **first** node is selected when you open a report. I will make this PR ready-for-review again when done.

PR #123, opened 2022-03-29
Test tab download upload

mhdirkse:


mhdirkse:
The tests fail because there is an error in ibis-ladybug, frank-runner, the test webapp or ladybug-frontend. @celmoussaoui, can you inform me if this has been fixed?

PR #322, opened 2023-12-22
feat: show multiple checkbox instead of button, default is show one

MatthijsSmets:
Remove button that toggles between show multiple and show one. Default is now show one, checkbox is added to show multiple

# How-to-demo

1) start the backend using the frank-runner on master branch
2) start the frontend on branch: feat/show-one-show-multiple-checkbox
3) switch the white box drop down to database storage
4) click on a row in the table
5) see that only one record can be shown in the debug file tree below the table
6) open the setting popup from the button iun the top left corner under the debug tab
7) click the setting: `Show multiple files in debug file-tree` to enable the checkbox
8) click on multiple rows in the table and see that they open in the debug tree below the table

## Notes
- Setting value is saved to local storage


Matthbo:
The  debug file tree keeps having multiple when I disable `Show multiple files in debug file-tree` again, this should probably clear the file tree so it reflects the changed setting

MatthijsSmets:
> The debug file tree keeps having multiple when I disable `Show multiple files in debug file-tree` again, this should probably clear the file tree so it reflects the changed setting

Fixed!

Matthbo (approved):

PR #324, opened 2024-01-15
Feat/create more space

MatthijsSmets:
Select all and deselect all buttons have been replaced with a checkbox in the table header.

Padding on table rows have been minimized to create more space.
Metadata table visible under the editor has been moved above it and can be shown or hidden with a button.
Table header names and column values have been shortened, hovering on the table header will still show the full name.

Table row spacing can be altered in settings, font size is also changed with the spacing.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -77,10 +77,10 @@
           "builder": "@angular-devkit/build-angular:dev-server",
           "configurations": {
             "production": {
-              "browserTarget": "ladybug:build:production"
+              "buildTarget": "ladybug:build:production"
             },
             "development": {
-              "browserTarget": "ladybug:build:development",
+              "buildTarget": "ladybug:build:development",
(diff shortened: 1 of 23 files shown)


angular.json
Matthbo, line 1: Im very curious what `buildTarget` is and why you changed to that instead of `browserTarget`. Ive never seen this before

MatthijsSmets: When running ng serve this is the warning i get:
image

Matthbo: Eh
image

Matthbo:
Currently spacing 0 is the default but I would change that to 1 as the default. 
If 1 is the default then people that want it small can make it smaller but it generally has better accessibility for people with to click on the right item

Matthbo (approved):

PR #330, opened 2024-01-18
Fix/commands in editor

MatthijsSmets:
Ctrl-f and other commands now work in the editor.
Search window is opened when editor is opened for the first time, this behaviour can be altered with the:
`Show search window in editor on load` setting in the settings modal

### DIFF
--- a/src/app/shared/components/editor/editor.component.ts
+++ b/src/app/shared/components/editor/editor.component.ts
@@ -1,21 +1,31 @@
-import { AfterViewInit, Component, ElementRef, ViewChild } from '@angular/core';
+import { AfterViewInit, Component, ElementRef, OnInit, ViewChild } from '@angular/core';
 import * as ace from 'ace-builds';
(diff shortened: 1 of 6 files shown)


src/app/shared/components/editor/editor.component.ts
```diff
@@ -1,20 +1,31 @@
...
+  ngOnInit(): void {
+    this.subscribeToSettings();
+  }
 
   ngAfterViewInit(): void {
     this.editor = ace.edit(this.editorElement.nativeElement);
-    ace.config.set('basePath', '../../../assets/editor-min');
+    console.log(ace.config.get('basePath'));
+    ace.config.set('basePath', 'assets/editor-min');
+    console.log(ace.config.get('basePath'));
```
Matthbo, line 28: Remove, console.log added for debug purposes

Matthbo (approved):

PR #347, opened 2024-02-12
chore: fix unnecessary shortening of values in table cells

MatthijsSmets:
Fix issue where values in table cells were always shortened if they exceeded certain length


src/app/shared/pipes/table-cell-shortener.pipe.ts
```diff
@@ -9,12 +11,10 @@ export class TableCellShortenerPipe implements PipeTransform {
     if (value == undefined) {
       return value;
     }
+    //Remove milliseconds if value is a timestamp with milliseconds
     if (value.match('\\b\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{3}\\b')) {
```
philipsens, line 15: It would be nice to declare the regex in a variable that describes what it is matching to.

src/app/custom-editor/custom-editor.component.ts
```diff
@@ -40,12 +40,16 @@ export class CustomEditorComponent implements OnInit, OnDestroy {
   editorChangesSubject: Subject<string> = new Subject<string>();
 
   @HostListener('window:keydown', ['$event'])
-  onSave(event: KeyboardEvent): void {
+  keyBoardListener(event: KeyboardEvent) {
```
Matthbo, line 43: didnt you always add a return type?

src/app/debug/table/table.component.html
```diff
@@ -68,16 +68,20 @@
           <th class="table-row-checkbox header-padding">
             <input data-cy-debug="selectAll" type="checkbox" class="vertical-middle" [checked]="allRowsSelected" (click)="selectAllRows()" title="Select all rows" [style.width]="getCheckBoxSize()" [style.height]="getCheckBoxSize()">
           </th>
-          <th class="text-center header-cell header-padding" *ngFor="let header of viewSettings.currentView.metadataLabels; let i = index"
-              mat-sort-header="{{i}}" [title]="header">{{getShortenedTableHeaderNames(header)}}</th>
+          <th class="text-center table-id-cell header-cell header-padding"
+              *ngFor="let header of viewSettings.currentView.metadataLabels; let i = index"
+              mat-sort-header="{{i}}"
```
Matthbo, line 73: Can't you do something like this? unless this doesn't work I'd rather see a non interpolated binding
```suggestion
              [mat-sort-header]="i"
```

MatthijsSmets: It doesnt accept an integer like that, it only accepts a string, which is why its implented like this.

src/app/shared/pipes/table-cell-shortener.pipe.ts
MatthijsSmets: Its only used in one place, so this comment should provide enought context.

src/app/custom-editor/custom-editor.component.ts
MatthijsSmets: Good point, fixed.

src/app/debug/table/table.component.html
Matthbo: Hm alright, you could've done `i+''` to make it a string

MatthijsSmets: Is there a notable difference between: `i+''` and `{{i}}` or just personal preference?

Matthbo: `{{i}}` interpolates it into the template and then binds it while `i+''` converts it to string and immediately binds it to the property without needing interpolation

MatthijsSmets: Yeah that does sound better, i'll correct it.

MatthijsSmets: This does seem to be allowed:
```suggestion
              [attr.mat-sort-header]="i"
```

src/app/shared/pipes/table-cell-shortener.pipe.ts
MatthijsSmets: But not less readable than: 
```js
    //Remove milliseconds if value is a timestamp with milliseconds
    if (value.match('\\b\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d{3}\\b')) {
```

philipsens: Is it explained? yes, is it readable code? no. 
If the code is complex enough that is needs a comment it is probably complex enough to get a method.

e.g. 
```typescript
removeMilisecondsFromTimestamp(value: string): string
```

MatthijsSmets: You're right, i've fixed it

philipsens (approved):

PR #348, opened 2024-02-13
Fix/compare tab switching nodes

MatthijsSmets:
Fix issue where switching between nodes in compare tree didnt update diff editor component


Matthbo (approved):

Matthbo (commented):
What did you test this with because comparing with database storage reports or simple report & another simple report doesnt update anything when I click on the different nodes

PR #367, opened 2024-03-07
Check copying report more strictly

mhdirkse:


cypress/e2e/testtab.cy.js
```diff
@@ -1,16 +1,23 @@
 describe('About the Test tab', () => {
-  beforeEach(() => {
-  });
-
   beforeEach(() => {
     cy.clearDebugStore();
     cy.createReport();
     cy.createOtherReport();
     cy.initializeApp();
+    let storageIds = [];
```
MatthijsSmets, line 7: Please make this const, the variable is not reassigned so it should not be let.

cypress.config.ts
```diff
@@ -15,5 +15,6 @@ export default defineConfig({
     viewportWidth: 1920,
     viewportHeight: 1080,
     experimentalRunAllSpecs: true,
+    video: true
```
MatthijsSmets, line 18: I don't see the added value of enabling video, we already know that the tests fail occasionally because they're instable. Enabling video will only make the tests slower.

MatthijsSmets (changes requested):

cypress.config.ts
mhdirkse: If the tests fail, you only have a screenshot where the assertion is not met. Video may be useful to investigate the root cause. We can turn of video again when the tests are stable again.

MatthijsSmets: But we already know why the tests fail, because they're unstable, so I don't see the added value of enabling video.

cypress/e2e/testtab.cy.js
mhdirkse: Gedaan.

MatthijsSmets (approved):

PR #371, opened 2024-03-12
Show reports in progress warning after meeting time threshold

MarkvdVorst:
before this merge:
- Merged PR on ladybug for new API get request  to save time threshold in TestTool (done)
- Merged PR on test-webapp that changes the value of time threshold in Ladybug TestTool object (done)

Added a feature that shows a warning when a report in progress has been going for longer than 5 minutes.

Added a cypress test that calls test-webapp to change time threshold of warning to one second for test and reverts it after.


src/app/debug/table/table.component.html
```diff
@@ -55,6 +55,10 @@
 
     </div>
 
+    @if (hasTimedOut) {
+      <h4 class="progress-report-timout">[One or more reports are in progress for more than {{ (this.reportsInProgressThreshold / 1000 / 60) }} minutes]</h4>
+    }
+
     <app-button [text]="'Compare'" [title]="'Compare'" *ngIf="showCompareButton()" (click)="compareTwoReports()"></app-button>
     <app-button [text]="'Open In New Tab'" *ngIf="showOpenInTabButton()" (click)="openReportInTab()"></app-button>
```
philipsens, line 63: No binding needed

src/app/debug/table/table.component.ts
```diff
@@ -215,10 +219,31 @@ export class TableComponent implements OnInit, OnDestroy {
...
         this.tableSettings.reportsInProgress = settings.reportsInProgress;
         this.tableSettings.estimatedMemoryUsage = settings.estMemory;
+        this.loadReportInProgressDates();
       },
     });
   }
 
+  loadReportInProgressDates() {
+    let hasChanged: boolean = false;
+    for (let i = 1; i <= this.tableSettings.reportsInProgress; i++) {
```
philipsens, line 229: Can't you rather use a for of loop?
I see that u need an index, but thats still possible.


```typescript
for (let [index, value] of myArray.entries()) {
  // your code goes here    
}
```

or 
```typescript
myArray.forEach((value, index) => {
  // your code goes here  
});
```

src/app/debug/table/table.component.ts
```diff
@@ -215,10 +219,31 @@ export class TableComponent implements OnInit, OnDestroy {
...
+      this.httpService.getReportInProgress(i).subscribe((report: Report) => {
+        if (this.reportsInProgress[report.correlationId] == null) {
+          this.reportsInProgress[report.correlationId] = report.startTime;
+        } else if (
+          Date.now() - new Date(this.reportsInProgress[report.correlationId]).getTime() >
+          this.reportsInProgressThreshold
```
philipsens, line 235: It might be nice to encapsulate this `if` to make it better readable what it is.

e.g. `if(this.reportIsInTreshosld(report))`

src/app/debug/table/table.component.ts
MarkvdVorst: Could you clarify the use of a for of/for each loop in this scenario? because the variable this.tableSettings.reportsInProgress is a number.

philipsens: When using a `for .. of` loop, each iteration of the iterable will be made available in a variable. You won't need to use `this.reportsInProgress[report.correlationId]` anymore. When using a destruct, you can also get an index.

src/app/debug/table/table.component.ts
```diff
@@ -226,14 +226,11 @@ export class TableComponent implements OnInit, OnDestroy {
 
   loadReportInProgressDates() {
     let hasChanged: boolean = false;
-    for (let i = 1; i <= this.tableSettings.reportsInProgress; i++) {
-      this.httpService.getReportInProgress(i).subscribe((report: Report) => {
+    for (let index = 1; index <= this.tableSettings.reportsInProgress; index++) {
+      this.httpService.getReportInProgress(index).subscribe((report: Report) => {
         if (this.reportsInProgress[report.correlationId] == null) {
           this.reportsInProgress[report.correlationId] = report.startTime;
```
philipsens:
```suggestion
        this.reportsInProgress[report.correlationId]  ??= report.startTime;
```

philipsens (changes requested):

src/app/debug/table/table.component.ts
philipsens: You told me in person that you are setting the values with this loop, and that the number updates separately.

src/app/debug/table/table.component.html
```diff
@@ -37,24 +37,28 @@
...
+          <input #openIndex type="number" min="0" value="{{tableSettings.numberOfReportsInProgress}}" max="{{tableSettings.numberOfReportsInProgress}}" (change)="disableReportInProgressButton(+openIndex.value, '#openReportInProgressButton')" class="form-control" data-cy-debug="openInProgressNo"/>
           <div class="input-group-append">
             <button data-cy-debug="openInProgress" class="btn btn-info" type="button" id="openReportInProgressButton" (click)="openReportInProgress(+openIndex.value)">Open</button>
           </div>
         </div>
           <div class="col input-group my-2 mx-1 px-2">
-            <input #deleteIndex type="number"  min="0" value="{{tableSettings.reportsInProgress}}" max="{{tableSettings.reportsInProgress}}" (change)="disableReportInProgressButton(+openIndex.value, '#deleteReportInProgressButton')" class="form-control" id="deleteInProgressNo"/>
+            <input #deleteIndex type="number" min="0" value="{{tableSettings.numberOfReportsInProgress}}" max="{{tableSettings.numberOfReportsInProgress}}" (change)="disableReportInProgressButton(+openIndex.value, '#deleteReportInProgressButton')" class="form-control" id="deleteInProgressNo"/>
```
Matthbo, line 50: Why do `max="{{tableSettings.numberOfReportsInProgress}}"` when you can make this a binding? Same goes for value
```
[value]="tableSettings.numberOfReportsInProgress" [max]=tableSettings.numberOfReportsInProgress
```

Always try to not interpolate what should be a binding.
Do pay attention to value since you need to use `[ngValue]` if its an object, but this is most likely only needed for `<select>` and its `<option>`s

src/app/debug/table/table.component.html
Matthbo: This means you can change it to just use interpolation since you're not really assigning a variable
```
<app-button [text]="'Compare'" [title]="'Compare'" ...></app-button>
```
Should become
```
<app-button text="Compare" title= "Compare" ...></app-button>
```

src/app/debug/table/table.component.ts
```diff
@@ -74,6 +74,9 @@ export class TableComponent implements OnInit, OnDestroy {
   defaultCheckBoxSize: number = 13;
   defaultFontSize: number = 8;
   fontSizeSpacingModifier: number = 1.2;
+  hasTimedOut: boolean = false;
+  reportsInProgress: { [key: string]: number } = {};
```
Matthbo, line 78: You can use the `Record<string, number>` type for this, its not mandatory but might be more readable and consistent with other projects that do use that type

src/app/shared/services/http.service.ts
```diff
@@ -85,6 +85,10 @@ export class HttpService {
       .pipe(catchError(this.handleError()));
   }
 
+  getReportsInProgressThresholdTime() {
+    return this.http.get<any>('api/testtool/in-progress/threshold-time').pipe(catchError(this.handleError()));
```
Matthbo, line 89: It hurts so much to see `any` here, are you sure you dont know the type?
Maybe something you and @MatthijsSmets can consider; using `unknown` instead of any so at least some form of type checking can be applied instead of having `any` which disables type checking for those objects

Matthbo (approved):
Left 1 comment, just a consideration

src/app/shared/services/http.service.ts
MarkvdVorst: I changed it to number

philipsens (approved):

src/app/debug/table/table.component.html
```diff
@@ -18,13 +18,13 @@
...
-    <app-button data-cy-debug="openSelected" [title]="'Open Selected Reports'" [text]="'Open Selected'" (click)="openSelected()"></app-button>
-    <app-button data-cy-debug="delete" [title]="'Delete'" [icon]="'fa fa-trash'" (click)="deleteSelected()"></app-button>
+    <app-button data-cy-debug="openSelected" title="Open Selected Reports" text="Open Selected" (click)="openSelected()"></app-button>
+    <app-button data-cy-debug="delete" title="Delete" icon="fa fa-trash" (click)="deleteSelected()"></app-button>
 
     <form (keydown.enter)="$event.preventDefault();" (keyup.enter)="changeTableLimit($event)">
       <div class='input-group'>
-          <input data-cy-debug='displayAmount' type='number' class='form-control input-field'
-                 value='{{tableSettings.displayAmount}}' (blur)='changeTableLimit($event)'>
+        <input data-cy-debug='displayAmount' type='number' class='form-control input-field'
```
mhdirkse, line 26: Ik zou dubbele quotes verwachten: `data-cy-debug="displayAmount"`. Dan kan je in Cypress testen refereren als: `cy.get('[data-cy-debug="displayAmount"]'`. Je hebt in src en in cypress dezelfde quotes om displayAmount heen.

src/app/debug/table/table.component.html
```diff
@@ -33,34 +33,38 @@
 
     <div class="my-2 mx-1 px-2">
       <select data-cy-change-view-dropdown class="form-control" (change)="changeView($event)">
-        <option [style.width]="viewDropdownBoxWidth" *ngFor="let view of viewSettings.views | keyvalue" [selected]="view.key === viewSettings.currentView.name" value="{{view.key}}">{{view.key}}</option>
+        <option [style.width]="viewDropdownBoxWidth" *ngFor="let view of viewSettings.views | keyvalue" [selected]="view.key === viewSettings.currentView.name" [value]="view.key">{{view.key}}</option>
       </select>
```
mhdirkse, line 37: Waarom verandert hier value={{ ... }} in [value]=" ... "? Heeft dit te maken met de functionaliteit die je wilt maken of is dit ongewild veranderd?

philipsens: Functioneel veranderd dit niet, het is eerder een "good practice" om altijd deze vorm van binding aan te nemen.

src/app/debug/table/table.component.ts
```diff
@@ -213,12 +217,35 @@ export class TableComponent implements OnInit, OnDestroy {
...
+        this.tableSettings.numberOfReportsInProgress = settings.reportsInProgress;
         this.tableSettings.estimatedMemoryUsage = settings.estMemory;
+        this.loadReportInProgressDates();
       },
     });
   }
 
+  loadReportInProgressDates() {
+    let hasChanged: boolean = false;
+    for (let index = 1; index <= this.tableSettings.numberOfReportsInProgress; index++) {
```
MatthijsSmets, line 229: Report with index 0 gets skipped?

src/app/shared/services/http.service.ts
```diff
@@ -85,6 +85,10 @@ export class HttpService {
       .pipe(catchError(this.handleError()));
   }
 
+  getReportsInProgressThresholdTime() {
```
MatthijsSmets, line 88: Please add the return type for this method

src/app/debug/table/table.component.ts
```diff
@@ -411,7 +438,7 @@ export class TableComponent implements OnInit, OnDestroy {
 
   disableReportInProgressButton(index: number, selector: string): void {
     let element: HTMLButtonElement = document.querySelector(selector)!;
```
MatthijsSmets, line 440: Please refactor this to the angular way.

MatthijsSmets (changes requested):
Looks good, just some minor adjustments

src/app/debug/table/table.component.ts
MarkvdVorst: this is done because in backend, the api does the following for some odd reason 
`Report report = testTool.getReportInProgress(index - 1);`

So that would have to be fixed first.

src/app/debug/table/table.component.ts
MarkvdVorst: removed it as it was unnecessary. ngIf statement was made in an element before in html file so it would have disabled the element anyway

src/app/shared/services/http.service.ts
MarkvdVorst: added return type

src/app/debug/table/table.component.ts
philipsens: In a way, it is logical. When asking "how many reports are in progress", 0 would mean there are none in progress. 
But thinking of it as an array filled with reports that are in progress, it would be more logical if the first item is 0. The length of the array is the amount of reports in progress.

src/app/shared/services/http.service.ts
philipsens: A lot of methods are missing a return type. This project as a whole needs to be refactored regarding its types (or the lack there off). I think you don't need to worry about it as much as long as there is a refactoring session in sight.

philipsens (approved):

MatthijsSmets (approved):

PR #373, opened 2024-03-22
Origin/refactor/setting correct types

Banshaan:
Fixed the getViews method so that the Observable is typed.

### DIFF
--- a/src/app/shared/services/http.service.ts
+++ b/src/app/shared/services/http.service.ts
@@ -32,8 +32,8 @@ export class HttpService {
     this.toastService.showSuccess(message);
   }
 
-  getViews(): Observable<{[key: string]: View}> {
-    return this.http.get<{[key: string]: View}>('api/testtool/views').pipe(catchError(this.handleError()));
+  getViews(): Observable<{ [key: string]: View }> {
+    return this.http.get<{ [key: string]: View }>('api/testtool/views').pipe(catchError(this.handleError()));
   }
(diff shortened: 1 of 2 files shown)


src/app/shared/services/http.service.ts
```diff
@@ -31,8 +32,8 @@ export class HttpService {
     this.toastService.showSuccess(message);
   }
 
-  getViews(): Observable<any> {
-    return this.http.get('api/testtool/views').pipe(catchError(this.handleError()));
+  getViews(): Observable<{ [key: string]: View }> {
+    return this.http.get<{ [key: string]: View }>('api/testtool/views').pipe(catchError(this.handleError()));
```
philipsens, line 36: Hi, thanks for the PR. 

Could you change `{ [key: string]: View }` to `Record<string, View>` that reads a bit better?

mhdirkse: Or Map<string, View>?. Wouldn't that be better, Sergi?

philipsens: Map does more than what @Banshaan needs it for in this case. `Record` is a Utility type: https://www.typescriptlang.org/docs/handbook/utility-types.html

Banshaan: Changed { [key: string]: View } to Record<string, View>

philipsens (approved):

PR #380, opened 2024-04-10
Origin/refactor/correct type

Banshaan:
Incoming backend data has now been given correct types in the http.service.ts inside the folder src/app/shared/services.

### DIFF
--- a/src/app/debug/table/table-settings-modal/table-settings-modal.component.ts
+++ b/src/app/debug/table/table-settings-modal/table-settings-modal.component.ts
@@ -133,11 +133,7 @@ export class TableSettingsModalComponent implements OnDestroy {
     if (localStorage.getItem('transformationEnabled')) {
       this.settingsForm.get('transformationEnabled')?.setValue(localStorage.getItem('transformationEnabled') == 'true');
     }
-
-    this.httpService.getTransformation(false).subscribe((response) => {
-      console.log(typeof response, response);
-      this.settingsForm.get('transformation')?.setValue(response.transformation);
-    });
+    this.httpService.getTransformation(false).subscribe((response) => this.settingsForm.get('transformation')?.setValue(response.transformation));
   }
 
   saveResponseSetting(response: any) {
--- a/src/app/debug/table/table-settings-modal/table-settings-modal.component.ts
+++ b/src/app/debug/table/table-settings-modal/table-settings-modal.component.ts
@@ -133,7 +133,9 @@ export class TableSettingsModalComponent implements OnDestroy {
     if (localStorage.getItem('transformationEnabled')) {
       this.settingsForm.get('transformationEnabled')?.setValue(localStorage.getItem('transformationEnabled') == 'true');
     }
-    this.httpService.getTransformation(false).subscribe((response) => this.settingsForm.get('transformation')?.setValue(response.transformation));
+    this.httpService
+      .getTransformation(false)
+      .subscribe((response) => this.settingsForm.get('transformation')?.setValue(response.transformation));
   }
 
   saveResponseSetting(response: any) {
--- a/src/app/debug/table/table-settings-modal/table-settings-modal.component.ts
+++ b/src/app/debug/table/table-settings-modal/table-settings-modal.component.ts
@@ -105,7 +105,7 @@ export class TableSettingsModalComponent implements OnDestroy {
     localStorage.setItem('transformationEnabled', form.transformationEnabled.toString());
     this.httpService.postTransformation(form.transformation).subscribe();
     const generatorEnabled: string = String(form.generatorEnabled === 'Enabled');
-    let data: any = {
+    let data: Object = {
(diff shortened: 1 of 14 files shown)


src/app/debug/table/table-settings-modal/table-settings-modal.component.ts
```diff
@@ -105,7 +105,7 @@ export class TableSettingsModalComponent implements OnDestroy {
     localStorage.setItem('transformationEnabled', form.transformationEnabled.toString());
     this.httpService.postTransformation(form.transformation).subscribe();
     const generatorEnabled: string = String(form.generatorEnabled === 'Enabled');
-    let data: any = {
+    let data: Object = {
```
Matthbo, line 108: This is technically correct, but wouldnt it be nice to know what's in the object type wise?
This goes for everything that uses the `Object` type in this pr

Banshaan: i'll work on fixing that

Banshaan: i changed all the Objects that i could find. They should be more descriptive than just "object" now. If i missed any, just let me know.

src/app/debug/table/table.component.ts
```diff
@@ -536,6 +536,8 @@ export class TableComponent implements OnInit, OnDestroy {
   }
 
   loadReportInProgressThreshold() {
-    this.httpService.getReportsInProgressThresholdTime().subscribe((time) => (this.reportsInProgressThreshold = time));
+    this.httpService.getReportsInProgressThresholdTime().subscribe((time) => {
```
MatthijsSmets, line 539:
```suggestion
    this.httpService.getReportsInProgressThresholdTime().subscribe((time: number) => {
```

MatthijsSmets (changes requested):
Looks good, I just have a few comments.

src/app/shared/services/http.service.ts
```diff
@@ -53,52 +60,52 @@ export class HttpService {
...
-  getLatestReports(amount: number, storage: string): Observable<any> {
+  getLatestReports(amount: number, storage: string): Observable<Report[]> {
     return this.http
-      .get<any>('api/report/latest/' + storage + '/' + amount)
+      .get<Report[]>('api/report/latest/' + storage + '/' + amount)
       .pipe(tap(() => this.handleSuccess('Latest' + amount + 'reports opened!')))
       .pipe(catchError(this.handleError()));
   }
 
   getReportInProgress(index: number) {
```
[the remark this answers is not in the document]
Banshaan: everything should now have a return type an i changed the let to const

MatthijsSmets (approved):

src/app/shared/services/http.service.ts
```diff
@@ -110,47 +117,46 @@ export class HttpService {
...
-  getReports(reportIds: string[], storage: string) {
+  getReports(reportIds: string[], storage: string): Observable<Record<string, CompareReport>> {
     return this.http
-      .get<any>(
-        'api/report/' + storage + '/?xml=true&globalTransformer=' + localStorage.getItem('transformationEnabled'),
-        { params: { storageIds: reportIds } },
-      )
+      .get<
+        Record<string, CompareReport>
+      >('api/report/' + storage + '/?xml=true&globalTransformer=' + localStorage.getItem('transformationEnabled'), { params: { storageIds: reportIds } })
```
philipsens, line 124: Could you, instead of using sting concatenation, use a template literal?
For example:
```suggestion
      >(`api/report/${storage}/?xml=true&globalTransformer=${localStorage.getItem('transformationEnabled')}, { params: { storageIds: reportIds } })
```

philipsens, line 124: It is always possible that a cookie is empty, does this cause weird stuff when calling the endpoint? If it does, you could add a check.

src/app/test/test.component.ts
```diff
@@ -252,10 +253,8 @@ export class TestComponent implements OnInit {
       if (!path.startsWith('/')) {
         path = '/' + path;
       }
-      let map = { path: path, action: action };
-      this.httpService.updatePath(reportIds, this.currentView.storageName, map).subscribe(() => {
-        this.loadData(path);
-      });
+      let map: UpdatePathSettings = { path: path, action: action };
```
philipsens, line 256: If the name and the variable are the same, it is possible to do this.

```suggestion
      let map: UpdatePathSettings = { path, action };
```

Looks cool right! 

This is however harder to refactor (if you rename one of the variables), so it's fine like how you did it.

Matthbo:
@Banshaan did you not see the added comments?

philipsens:
I only later saw that my comments were on code that you didn't write. But it any case, it is nice to fix stuff like this the moment we spot it.

In the future, resolve all conversations before merging. If you don't agree with some comment or review, you can just say so, but then we know your perspective.

PR #390, opened 2024-04-23
Added a new filter-side-drawer component for filtering reports

MarkvdVorst:
Pull Request 231 of the Ladybug project should be accepted before accepting this pull request.

Added a filter side drawer that replaces the old filter feature.
The new filter component now communicates with a filter service to send and receive the filters of the table.

Most of the functionality in the table component remains the same to filter the table. Parts that require the input of the filter however have been changed so that it now communicates with the filter service.

This feature also solves issue 290; When multiple filters are filled only one is used.

### DIFF
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
@@ -1,0 +1,21 @@
+.vertical-panel {
+    position: fixed;
+    top: 0;
+    right: 0;
+    height: 100vh;
+    width: 300px;
+    background-color: #f0f0f0;
+    box-shadow: 0 4px 8px rgba(0, 0, 0, 0.1);
+    z-index: 1000;
+}
+
+.vertical-panel * {
+  margin-top: 1rem;
+  margin-left: 1rem;
+}
+
+.close-filter-btn {
+  position: absolute;
+  bottom: 1rem;
+  left: 1rem;
+}
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
@@ -1,4 +1,4 @@
-.vertical-panel {
+.filter-vertical-panel {
   position: fixed;
   top: 0;
   right: 0;
@@ -16,6 +16,13 @@
   position: inherit;
   bottom: 1rem;
   margin-left: 1rem;
+}
+
+.clear-filter-btn {
+  position: inherit;;
+  bottom: 1rem;
+  margin-right: 1rem;
+  right: 0;
 }
 
 .filter-input-container {
@@ -46,6 +53,8 @@
 
 .filter-label {
   font-size: 1rem;
+  margin-left: 1rem;
+  margin-top: 1rem;
 }
 
 .metadata-input-label {
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
@@ -2,30 +2,54 @@
   position: fixed;
   top: 0;
   right: 0;
-  height: 100vh;
+  height: 100%;
   width: 300px;
   background-color: #f0f0f0;
   box-shadow: 0 0 8px 1px gray;
   z-index: 1000;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
 }
 
 .close-filter-btn {
-  position: absolute;
+  position: inherit;
   bottom: 1rem;
-  left: 1rem;
+  margin-left: 1rem;
 }
 
-.metadata-input-container {
+.filter-input-container {
   display: flex;
   flex-direction: column;
   justify-content: space-evenly;
-  height: 80vh;
+  align-items: flex-start;
+  height: 90%;
+  width: 100%;
+  margin-left: 1rem;
+}
+
+.metadata-input-container {
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
 }
 
 .metadata-input {
   padding: 5px;
+  width: 80%;
+  border: 1px solid #ced4da;
+  border-radius: .25rem;
+  background-clip: padding-box;
+  height: calc(1.5em + .75rem + 2px);
 }
 
 .filter-label {
   font-size: 1rem;
+}
+
+.metadata-input-label {
+  display: flex;
+  flex-direction: column;
+  font-weight: bold;
 }
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
@@ -9,13 +9,23 @@
   z-index: 1000;
 }
 
-.vertical-panel * {
-  margin-top: 1rem;
-  margin-left: 1rem;
-}
-
 .close-filter-btn {
   position: absolute;
   bottom: 1rem;
   left: 1rem;
+}
+
+.metadata-input-container {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-evenly;
+  height: 80vh;
+}
+
+.metadata-input {
+  padding: 5px;
+}
+
+.filter-label {
+  font-size: 1rem;
 }
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.css
@@ -1,12 +1,12 @@
 .vertical-panel {
-    position: fixed;
-    top: 0;
-    right: 0;
-    height: 100vh;
-    width: 300px;
-    background-color: #f0f0f0;
-    box-shadow: 0 4px 8px rgba(0, 0, 0, 0.1);
-    z-index: 1000;
+  position: fixed;
+  top: 0;
+  right: 0;
+  height: 100vh;
+  width: 300px;
+  background-color: #f0f0f0;
+  box-shadow: 0 0 8px 1px gray;
+  z-index: 1000;
 }
 
 .vertical-panel * {
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
@@ -1,12 +1,11 @@
 <div @removeTrigger class="vertical-panel" *ngIf="shouldShowFilter">
-  <p>wow</p>
-  <p>wow</p>
-  <p>wow</p>
-  <p>wow</p>
-  <p>wow</p>
-  <p>wow</p>
-  <p>wow</p>
-  <p>wow</p>
-  <p>wow</p>
+  <label class="filter-label">Filter Reports</label>
+  <div class="metadata-input-container">
+    <input *ngFor="let metadataName of metadataNames"
+           class="metadata-input"
+           type="text"
+           [title]="metadataName"
+           placeholder="Filter for {{metadataName}}">
+  </div>
   <button class="close-filter-btn btn btn-info my-2 mx-1 px-3" (click)="closeFilter()">Close</button>
 </div>
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
@@ -1,5 +1,5 @@
 <div @removeTrigger class="filter-vertical-panel" data-cy-debug="filter-side-drawer" *ngIf="shouldShowFilter">
-  <label class="filter-label">Filter Reports</label>
+  <label class="filter-label">Filter</label>
   <div class="filter-input-container">
     @for (metadataName of metadataNames; track metadataNames) {
       <div class="metadata-input-container">
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
@@ -1,0 +1,12 @@
+<div class="vertical-panel" *ngIf="shouldShowFilter">
+  <p>wow</p>
+  <p>wow</p>
+  <p>wow</p>
+  <p>wow</p>
+  <p>wow</p>
+  <p>wow</p>
+  <p>wow</p>
+  <p>wow</p>
+  <p>wow</p>
+  <button class="close-filter-btn btn btn-info my-2 mx-1 px-3" (click)="closeFilter()">Close</button>
+</div>
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
@@ -1,4 +1,4 @@
-<div @removeTrigger class="vertical-panel" *ngIf="shouldShowFilter">
+<div @removeTrigger class="filter-vertical-panel" data-cy-debug="filter-side-drawer" *ngIf="shouldShowFilter">
   <label class="filter-label">Filter Reports</label>
   <div class="filter-input-container">
     @for (metadataName of metadataNames; track metadataNames) {
@@ -7,11 +7,14 @@
         <input
           class="metadata-input"
           type="text"
+          [(ngModel)]="filters[metadataName]"
           [title]="metadataName"
           placeholder="Filter for {{metadataName}}"
-          (keydown.enter)="updateFilter($event, metadataName)">
+          (keydown.enter)="updateFilter($event, metadataName)"
+          data-cy-debug="tableFilter">
       </div>
     }
   </div>
-  <button class="close-filter-btn btn btn-info" (click)="closeFilter()">Close</button>
+  <button class="close-filter-btn btn btn-info" (click)="closeFilter()" data-cy-debug="close-filter-btn">Close</button>
+  <button class="clear-filter-btn btn btn-info" (click)="resetFilter()" data-cy-debug="clear-filter-btn">Clear</button>
 </div>
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
@@ -1,4 +1,4 @@
-<div class="vertical-panel" *ngIf="shouldShowFilter">
+<div @removeTrigger class="vertical-panel" *ngIf="shouldShowFilter">
   <p>wow</p>
   <p>wow</p>
   <p>wow</p>
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.html
@@ -1,11 +1,17 @@
 <div @removeTrigger class="vertical-panel" *ngIf="shouldShowFilter">
   <label class="filter-label">Filter Reports</label>
-  <div class="metadata-input-container">
-    <input *ngFor="let metadataName of metadataNames"
-           class="metadata-input"
-           type="text"
-           [title]="metadataName"
-           placeholder="Filter for {{metadataName}}">
+  <div class="filter-input-container">
+    @for (metadataName of metadataNames; track metadataNames) {
+      <div class="metadata-input-container">
+        <label class="metadata-input-label">{{ metadataName }}</label>
+        <input
+          class="metadata-input"
+          type="text"
+          [title]="metadataName"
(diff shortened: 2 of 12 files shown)


src/app/debug/filter-side-drawer/filter-side-drawer.component.html
```diff
@@ -0,0 +1,20 @@
...
+  <div class="filter-input-container">
+    @for (metadataName of metadataNames; track metadataNames) {
+      <div class="metadata-input-container">
+        <label class="metadata-input-label">{{ metadataName }}</label>
+        <input
+          class="metadata-input"
+          type="text"
+          [(ngModel)]="filters[metadataName]"
+          [title]="metadataName"
+          placeholder="Filter for {{metadataName}}"
```
philipsens, line 12: We created a convention (which is not published yet) for sting interpolation and binding:

```typescript
// Bad
<option value="{{view.key}}">
  {{ view.value }}
</option>


// Bad
<option [value]="'{{view.key}}px'">
  {{ view.value }}
</option>


// Good
<option [value]="view.key + 'px'">
  {{ view.value }}
</option>


// Good
<option value="15px">
  {{ view.value }}
</option>
```

So then your code should be:

```suggestion
          [placeholder]="'Filter for' + metadataName"
```

MarkvdVorst: alright. good to know

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -0,0 +1,66 @@
...
+    ]),
+  ],
+})
+export class FilterSideDrawerComponent implements OnDestroy, OnInit {
+  protected shouldShowFilter!: boolean;
+  shouldShowFilterSubscriber!: Subscription;
+  protected metadataNames!: string[];
+  metadataNamesSubscriber!: Subscription;
+  protected filters: Record<string, string> = {};
+  filterSubscriber!: Subscription;
```
philipsens, line 27: I think its a bit nicer to order the class members by accessibility

```suggestion
  protected shouldShowFilter!: boolean;
  protected metadataNames!: string[];
  protected filters: Record<string, string> = {};

  shouldShowFilterSubscriber!: Subscription;
  metadataNamesSubscriber!: Subscription;
  filterSubscriber!: Subscription;
```

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.shouldShowFilterSubscriber.unsubscribe();
+    this.metadataNamesSubscriber.unsubscribe();
+    this.filterSubscriber.unsubscribe();
+  }
+
+  closeFilter() {
+    this.filterService.setShowFilter(false);
+  }
+
+  protected readonly auto = auto;
```
philipsens, line 57: Where does this come from?

src/app/debug/filter-side-drawer/filter.service.ts
```diff
@@ -0,0 +1,35 @@
...
+export class FilterService {
+  private showFilterSubject: Subject<boolean> = new Subject();
+  showFilterObserver: Observable<boolean> = this.showFilterSubject.asObservable();
+
+  private metadataNamesSubject: Subject<string[]> = new Subject();
+  metadataNamesObserver: Observable<string[]> = this.metadataNamesSubject.asObservable();
+
+  private filters: Record<string, string> = {};
+  private filterContextSubject: Subject<Record<string, string>> = new Subject();
+  filterContextObserver: Observable<Record<string, string>> = this.filterContextSubject.asObservable();
```
philipsens, line 16: Here to. Might even add a $ to the observers, IDK what @Matthbo thinks about that. I'll add it to the conventions.

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
MarkvdVorst: yeah that makes sense, I'll add the change

src/app/debug/filter-side-drawer/filter.service.ts
Matthbo: In RxJS and Angular docs its common to indicate a subject by either a normal variable name or with the suffix "Subject".
Then the Observable version of that subject will have the normal variable name with the suffix "$" instead.
```ts
const booleanVarSubject: Subject<boolean> = new Subject(false);
const booleanVar$: Observable<boolean> = booleanVarSubject.asObservable();
```
To me this seems very clear since the subject can do everything while the observable is only meant for public accessibility and can only be subscribed to

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
MarkvdVorst: must've accidentally added that with a quick solution thing. it's not supposed to be there. thanks for pointing it out

src/app/debug/filter-side-drawer/filter-side-drawer.component.css
MatthijsSmets, line 1: On smaller screen sizes not all filter fields are visible, it would be nice to be able to scroll.

MatthijsSmets (changes requested):

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  updateFilter(filterEvent: any, metadataName: string): void {
```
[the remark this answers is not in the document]
MarkvdVorst: angular only accepts Event as the type. So I casted it to the following
`updateFilter(filterEvent: Event, metadataName: string): void {
    this.filterService.updateFilterContext(metadataName, (filterEvent.target as HTMLInputElement).value);
  }`

src/app/debug/table/table.component.html
```diff
@@ -79,31 +80,6 @@ <h4 class="progress-report-timout">[One or more reports are in progress for more
...
-        <tr data-cy-debug="tableFilterRow" id="filterRow" *ngIf="tableSettings.showFilter">
-          <th class="vertical-middle">
-            <app-button icon="fa fa-undo" title="Clear filters" (click)="clearFilters()"></app-button>
-          </th>
-          <th *ngFor="let header of viewSettings.currentView.metadataNames" class="filter-header">
-            <div class="input-group">
-              <div>
-                <input
-                  data-cy-debug="tableFilter"
-                  [matAutocomplete]="auto"
```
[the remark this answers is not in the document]
MarkvdVorst: I implemented the mat-autocmplete feature and fixed one of the bugs that I noticed on live version. Not all options were being visible or usable. This was due to an oversight in the table component.
Apart from that I added an empty option for each of the filters so a user can more easily reset a specific field for the filter

src/app/debug/table/table.component.ts
```diff
@@ -375,15 +390,21 @@ export class TableComponent implements OnInit, OnDestroy {
...
 
-  changeFilter(event: any, header: string, value?: string): void {
-    const filterValue: string = value ?? event.target.value;
-    if (this.currentFilters.get(header) !== filterValue) {
-      this.tableSettings.filterValue = filterValue ?? '.*';
-      this.tableSettings.filterHeader = filterValue;
-      this.tableSettings.filterHeader &&= header;
+  changeFilter(filters: Record<string, string>): void {
+    if (Object.keys(filters).length === 0) {
+      this.tableSettings.filterValues = [];
```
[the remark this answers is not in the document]
MarkvdVorst: I removed the resetFilter method in the filter sidedrawer component and now it just directly calls the reset filter in the filter service

src/app/debug/filter-side-drawer/filter-side-drawer.component.css
MarkvdVorst: I fixed the issue with a flexbox, it should work on all sizes now

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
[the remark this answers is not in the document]
MatthijsSmets: On second thought, seeing that you already use ngModel, have a look at [ngModelChange](https://angular.io/api/forms/NgModel#properties)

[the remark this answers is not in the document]
MarkvdVorst: The problem with that currently with how the program works is that it just calls the API over and over again and then we get this on the screen for every letter typed
image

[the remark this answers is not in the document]
MarkvdVorst: I did change the type to just string now so the actual data can be send directly to the updateFilter() method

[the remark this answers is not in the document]
MarkvdVorst: @MatthijsSmets I added the debounceTime to prevent it. Thanks for the tip

MatthijsSmets:
image
When there are no values the mat-autocomplete shows a dropdown with an empty suggestion.

MarkvdVorst:
> image When there are no values the mat-autocomplete shows a dropdown with an empty suggestion.

yes that's supposed to be there. that option will also be there when there there are other options. it's so that a user can more easily empty a filter category. or do you mean that this shows even when there are no reports at all?

src/app/debug/table/table.component.ts
```diff
@@ -488,18 +509,20 @@ export class TableComponent implements OnInit, OnDestroy {
...
       for (let element of data) {
-        uniqueValues.add(element[lowerHeaderName]);
-        uniqueValues.add(element[upperHeaderName]);
+        if (element[lowerHeaderName] != undefined) uniqueValues.add(element[lowerHeaderName]);
+        if (element[upperHeaderName] != undefined) uniqueValues.add(element[upperHeaderName]);
+        if (element[headerName] != undefined) uniqueValues.add(element[headerName]);
       }
       this.tableSettings.uniqueValues.set(
         lowerHeaderName,
         uniqueValues.size < 15 ? this.sortUniqueValues(uniqueValues) : ([] as string[]),
```
MatthijsSmets, line 523: I know you didn't write this, but could you remove this magic number?

src/app/debug/filter-side-drawer/filter-side-drawer.component.css
```diff
@@ -0,0 +1,59 @@
...
+  box-shadow: 0 0 8px 1px gray;
+  z-index: 1000;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+}
+
+.filter-input-container {
+  margin-left: 1rem;
+  margin-right: 1rem;
```
MatthijsSmets, line 17: This can be shortened:
```suggestion
  margin: 0 1rem;
```

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  shouldShowFilterSubscriber!: Subscription;
+  metadataNamesSubscriber!: Subscription;
+  filterSubscriber!: Subscription;
+  currentRecordsSubscriber!: Subscription;
+
+  constructor(protected filterService: FilterService) {}
+
+  ngOnInit(): void {
+    this.shouldShowFilterSubscriber = this.filterService.showFilter$.subscribe((show): void => {
```
MatthijsSmets, line 34: Add a type to show, just like you do with the rest of the subscriptions

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -0,0 +1,66 @@
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import { FilterService } from './filter.service';
+import { Subscription } from 'rxjs';
+import { animate, style, transition, trigger } from '@angular/animations';
+
+@Component({
```
MatthijsSmets, line 6: Please make this component standalone, have a look at the [docs](https://angular.io/guide/standalone-components). Don't forget to remove the declaration of the component in the app.module.ts

src/app/debug/filter-side-drawer/filter-side-drawer.component.html
```diff
@@ -0,0 +1,29 @@
+<div @removeTrigger class="filter-vertical-panel" data-cy-debug="filter-side-drawer" *ngIf="shouldShowFilter">
```
MatthijsSmets, line 1: Please replace this `*ngIf` with `@if`, just like you do with the `@for`

src/app/debug/filter-side-drawer/filter.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class FilterService {
+  private showFilterSubject: Subject<boolean> = new Subject();
+  private metadataNamesSubject: Subject<string[]> = new Subject();
+  private filterContextSubject: Subject<Record<string, string>> = new Subject();
+  private currentRecordsSubject: Subject<Map<string, Array<string>>> = new Subject();
+  private filters: Record<string, string> = {};
```
MatthijsSmets, line 12: Is there a reason that it's initialized with an empty object here?

MatthijsSmets (changes requested):

MatthijsSmets:
> > image When there are no values the mat-autocomplete shows a dropdown with an empty suggestion.
> 
> yes that's supposed to be there. that option will also be there when there are other options. it's so that a user can more easily empty a filter category. or do you mean that this shows even when there are no reports at all? if you think this is unnecessary I can remove it.
> 
> I could also just change it so it doesn't show when the amount of reports on display is zero

Yeah it would be nice if it didnt show an empty option and to allow the user to set an empty option with a button.
Something like [this](https://getbootstrap.com/docs/4.0/components/input-group/#basic-example):
```html
      <div class="input-container">
        <label class="metadata-input-label">{{ metadataName }}</label>
        <div class="input-group">
          <input
            class="metadata-input form-control"
            type="text"
            [title]="metadataName"
            [(ngModel)]="filterService.getCurrentFilterContext()[metadataName]"
            [matAutocomplete]="auto"
            [placeholder]="'Filter for ' + metadataName"
            (ngModelChange)="updateFilter($event, metadataName)"
            data-cy-debug="tableFilter">
          <div class="input-group-append">
            <button class="btn btn-sm btn-danger button-group-text">Clear</button>
          </div>
        </div>
        <mat-autocomplete #auto="matAutocomplete" (optionSelected)="updateFilter($event.option.value, metadataName)">
          @for (option of currentRecords.get(metadataName.toLocaleLowerCase()); track currentRecords) {
            <mat-option [value]="option" class="filter-option" data-cy-debug="matAutocompleteOption">{{ option }}</mat-option>
          }
        </mat-autocomplete>
      </div>
```

MatthijsSmets:
Right now the focus ring is cut off:
image
because of the container width, adding some padding to the container will fix this.

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -0,0 +1,66 @@
...
+        this.currentRecords = records;
+      },
+    );
+  }
+
+  ngOnDestroy(): void {
+    this.shouldShowFilterSubscriber.unsubscribe();
+    this.metadataNamesSubscriber.unsubscribe();
+    this.filterSubscriber.unsubscribe();
+    this.currentRecordsSubscriber.unsubscribe();
```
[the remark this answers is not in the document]
MarkvdVorst: also added this in the new commit

src/app/debug/filter-side-drawer/filter.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  showFilter$: Observable<boolean> = this.showFilterSubject.asObservable();
+  metadataNames$: Observable<string[]> = this.metadataNamesSubject.asObservable();
+  filterContext$: Observable<Record<string, string>> = this.filterContextSubject.pipe(debounceTime(300));
+  currentRecords$: Observable<Map<string, Array<string>>> = this.currentRecordsSubject.asObservable();
```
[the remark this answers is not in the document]
MarkvdVorst: not really. it was an oversight. They're both Maps now

src/app/debug/table/table.component.ts
```diff
@@ -88,17 +91,27 @@ export class TableComponent implements OnInit, OnDestroy {
...
+      },
+    );
     this.loadData();
     this.listenForViewUpdate();
     this.subscribeToSettingsObservables();
   }
 
   ngOnDestroy(): void {
     this.tableSpacingSubscription.unsubscribe();
+    this.filterContextSubscription.unsubscribe();
```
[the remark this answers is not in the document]
MarkvdVorst: added this in the new commit

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
MarkvdVorst: I added the :void to the shouldShowFilterSubscriber subscription

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
MarkvdVorst: FilterSideDrawer is now standalone

src/app/debug/filter-side-drawer/filter-side-drawer.component.html
MarkvdVorst: added the change

src/app/debug/filter-side-drawer/filter.service.ts
MarkvdVorst: since I've changed the records all into Map types, this was no longer necessary.

src/app/debug/table/table.component.ts
```diff
@@ -490,18 +521,27 @@ export class TableComponent implements OnInit, OnDestroy {
...
+        }
+        if (element[headerName]) {
+          uniqueValues.add(element[headerName]);
+        }
       }
+      const MAX_AMOUNT_OF_FILTER_SUGGESTIONS: number = 15;
       this.tableSettings.uniqueValues.set(
         lowerHeaderName,
-        uniqueValues.size < 15 ? this.sortUniqueValues(uniqueValues) : ([] as string[]),
+        uniqueValues.size < MAX_AMOUNT_OF_FILTER_SUGGESTIONS ? this.sortUniqueValues(uniqueValues) : ([] as string[]),
```
MatthijsSmets, line 544:
```suggestion
        uniqueValues.size < MAX_AMOUNT_OF_FILTER_SUGGESTIONS ? this.sortUniqueValues(uniqueValues) : [],
```

src/app/debug/filter-side-drawer/filter-side-drawer.component.html
```diff
@@ -0,0 +1,55 @@
...
+                  (click)="removeFilter(metadataName)"
+                  data-cy-debug="filterClearButton">
+                  Clear
+                </button>
+              </div>
+            </div>
+            <mat-autocomplete
+              #auto="matAutocomplete"
+              (optionSelected)="updateFilter($event.option.value, metadataName)">
+              @for (option of currentRecords.get(metadataName.toLocaleLowerCase()); track currentRecords) {
```
MatthijsSmets, line 31: Same here
```suggestion
              @for (option of currentRecords.get(metadataName.toLocaleLowerCase()); track option) {
```

src/app/debug/filter-side-drawer/filter-side-drawer.component.html
```diff
@@ -0,0 +1,55 @@
+@if (shouldShowFilter) {
+  <div @removeTrigger class="filter-vertical-panel" data-cy-debug="filter-side-drawer">
+    <div class="panel-container">
+      <h1 class="h1 filter-label">Filter</h1>
+      <div class="filter-input-container">
+        @for (metadataName of metadataNames; track metadataNames) {
```
MatthijsSmets, line 6: The track expression should be used to identify each element instead of the collection to increase performance, check [the docs](https://angular.io/api/core/for#description) for more info
```suggestion
        @for (metadataName of metadataNames; track metadataName) {
```

MatthijsSmets (changes requested):
Almost there!

MatthijsSmets (approved):

MatthijsSmets (approved):

MarkvdVorst:
@Matthbo @philipsens Does anything else need to be changed before merge?

Issue #393, opened 2024-04-30
Refactor updatePath method in test.component.ts


MatthijsSmets:
In the test component, update path and the form that triggers it should be refactored to use Angular's built in form support, instead of using query selector and value={{}}. This is not the only place where this happens all places where a querySelector is used can be replaced with the angular way.

MarkvdVorst:
The queryselector already gets fixed with the PR for issue #409

MarkvdVorst:
The use of value={{}} example gets fixed in the PR for issue #396

PR #395, opened 2024-05-01
Replace jqxTree in test component with ng-simple-file-tree

MatthijsSmets:
Replaced jqxTree in the test component with ng-simple-file-tree as preparation for completely removing jqwidgets-ng package


src/app/debug/table/table-settings-modal/table-settings-modal.component.html
```diff
@@ -113,7 +113,7 @@ <h5>Report table</h5>
       <button title="Revert changes" type="button" class="btn btn-info" (click)="loadSettings()"><i
         class="fa fa-undo"></i></button>
       <button data-cy-settings='saveChanges' title="Save changes" type="submit" id="saveTableSettings"
-              class="btn btn-info" (click)="modal.dismiss()"><i class="fa fa-save"></i></button>
+              class="btn btn-info" (click)="modal.dismiss(); saveSettings()"><i class="fa fa-save"></i></button>
```
Matthbo, line 116: Small clean improvement: move click logic to its own function (prevent multiple statements in template)

src/app/test/test-folder-tree/test-folder-tree.component.ts
```diff
@@ -1,86 +1,97 @@
...
-        currentFolders.push(currentFolder);
-      }
-
-      this.latestAddedFolder = currentFolder;
-      this.recursivelyAddFolders(folderNames, currentFolder.items, currentFolder.value);
+  setData(data: CreateTreeItem[]) {
+    this.resetTree();
+    this.originalItems = data;
+    const tempItems = this.convertToFolders(data);
+    let rootFolder = Object.assign({}, this.rootFolder);
```
Matthbo, line 51: Why a let? Its not reassigned right? 
Instead of `Object.assign({}, ...)` you could use the spread operator for shallow copying (unless this needs deep copying of course)

src/app/test/test-folder-tree/test-folder-tree.component.ts
```diff
@@ -1,86 +1,97 @@
...
-      if (item) {
-        this.treeReference.selectItem(item);
-        this.changeFolderEvent.next(item.value);
+  convertToFolders(items: CreateTreeItem[]) {
+    const tempItems: CreateTreeItem[] = [];
+    for (let item of items) {
+      if (item.path) {
+        try {
+          tempItems.push(this.createFolderFromPath(item.path));
+        } catch {}
```
Matthbo, line 64: Is an error thrown here? Seems like there is always an array so push should never fail.
Also an empty catch okay in some cases but in most cases you at least want to log something or do a different action, now its even more unclear what the error could be about

src/app/test/test-folder-tree/test-folder-tree.component.ts
```diff
@@ -1,86 +1,97 @@
...
+    }
+    if (index === -1) {
+      throw new Error('Path had only 1 slash');
+    }
+    let folderName = path.slice(0, index);
+    const remainingPath = path.replace(folderName, '');
+    if (folderName.startsWith('/')) {
+      folderName = folderName.slice(1);
+    }
+    const treeItem = { name: folderName, icon: 'assets/tree-icons/folder.svg' } as CreateTreeItem;
```
Matthbo, line 83:
```suggestion
    const treeItem: CreateTreeItem = { name: folderName, icon: 'assets/tree-icons/folder.svg' };
```

src/app/test/test.component.ts
```diff
@@ -30,18 +30,22 @@ export class TestComponent implements OnInit {
...
+  @ViewChild(TestFolderTreeComponent) testFileTreeComponent!: TestFolderTreeComponent;
 
   constructor(
     private httpService: HttpService,
     private helperService: HelperService,
     private toastService: ToastService,
     private tabService: TabService,
-  ) {}
+  ) {
+    this.getGeneratorStatus();
```
Matthbo, line 43: Could this be in OnInit?

src/app/test/test-folder-tree/test-folder-tree.component.ts
MatthijsSmets: You're right it should be const.

src/app/test/test.component.ts
MatthijsSmets: It can be, but it retrieves from localstorage so it's not necessary to call it in onInit

src/app/test/test-folder-tree/test-folder-tree.component.ts
MatthijsSmets: And yes this needs to be a deep copy, thats why i've used Objects.assign

src/app/test/test.component.ts
Matthbo: I could say the same for calling it in the constructor instead of using the correct lifecycle method

src/app/test/test.component.html
```diff
@@ -5,79 +5,93 @@
...
       </button>
     </div>
 
 
     <div class="form-group mt-2">
       <label for="moveToInput">Move/Copy to:</label>
       <form class="input-group">
-        <input type="text" class="form-control mr-1 my-2 py-0" id="moveToInput" name="filter" value="{{currentFilter}}"/>
+        <input type="text" class="form-control mr-1 my-2 py-0" id="moveToInput" name="filter"
+               value="{{currentFilter}}" />
```
philipsens, line 36: (Jaja niet jouw code)
Prefer binding over interpolation

src/app/custom-editor/custom-editor.component.ts
```diff
@@ -208,11 +208,13 @@ export class CustomEditorComponent implements OnInit, OnDestroy, OnChanges {
...
+    if (value) {
+      for (let i = 0; i < value.length; i++) {
+        if (value.charAt(i) === ' ' || value.charAt(i) === '\t') {
+          continue;
+        }
+        return value.charAt(i) === '<';
       }
-      return value.charAt(i) === '<';
     }
     return false;
```
philipsens, line 219: Maybe an early return would be nice here.

src/app/test/test.component.ts
MatthijsSmets: Yes however the method sets a class variable that is needed in the HTML, therefore getting it before the ngOnInit would be better as otherwise an `NG0100: Expression has changed after it was checked` error is thrown.

src/app/test/test.component.ts
```diff
@@ -246,25 +255,46 @@ export class TestComponent implements OnInit {
...
+    }
+  }
+
+  transformFilter(value: string): string {
+    if (value !== '' && value !== 'Reports') {
+      if (!value.startsWith('/')) {
+        value = `/${value}`;
+      }
+      if (!value.endsWith('/')) {
+        value = `${value}/`;
```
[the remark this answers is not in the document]
MatthijsSmets: Nice, I didn't know this is possible fixed it.

src/app/debug/table/table-settings-modal/table-settings-modal.component.html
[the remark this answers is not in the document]
MatthijsSmets: Done

src/app/test/test.component.ts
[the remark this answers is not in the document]
MatthijsSmets: I agree that it looks cleaner, but I think that in terms of readability the template string is better as its also used to prefix the string with a slash. Just to stay consistent 😉

src/app/test/test.component.ts
```diff
@@ -246,25 +255,46 @@ export class TestComponent implements OnInit {
...
     let reportIds: string[] = this.helperService.getSelectedIds(this.reports);
     if (reportIds.length > 0) {
-      let path = (document.querySelector('#moveToInput')! as HTMLInputElement).value;
-      if (!path.endsWith('/')) {
-        path = path + '/';
-      }
-      if (!path.startsWith('/')) {
-        path = '/' + path;
-      }
+      let path: string = (document.querySelector('#moveToInput')! as HTMLInputElement).value;
```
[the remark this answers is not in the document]
MatthijsSmets: Yeah this needs to be replaced with either on of the angular solutions for forms. I thought it'd be better to do all those changes in one pull request. An issue has been made: #393.

src/app/test/test.component.ts
```diff
@@ -246,25 +255,46 @@ export class TestComponent implements OnInit {
...
+      path = this.addSlashesToPathEnds(path);
       let map: UpdatePathSettings = { path: path, action: action };
```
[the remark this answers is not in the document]
MatthijsSmets: Fixed

src/app/test/test.component.ts
```diff
@@ -246,25 +255,46 @@ export class TestComponent implements OnInit {
...
       this.httpService.updatePath(reportIds, this.currentView.storageName, map).subscribe(() => this.loadData(path));
     } else {
       this.toastService.showWarning('No Report Selected!');
     }
   }
 
+  addSlashesToPathEnds(path: string): string {
+    if (path.length > 0 && !path.startsWith('/')) {
+      path = '/' + path;
```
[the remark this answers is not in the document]
MatthijsSmets: Yes maybe best to be consistent, I've changed it all to a template string.

src/app/test/test.component.ts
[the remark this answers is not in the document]
philipsens: I think there's a difference between string concatenation and creating a new string. But in this case, having these values so close together it might look nicer.

src/app/test/test.component.html
MatthijsSmets: This is the what the querySelector is used for that you've commented about earlier, so it will be fixed the same way.

src/app/test/test.component.html
```diff
@@ -5,79 +5,93 @@
...
-              </td>
-              <td>
+            </td>
+            <td>
                 <span *ngIf="getReranReport(report.storageId)">
-                  <app-button data-cy-test="replaceReport" [text]="'Replace'" [title]="'Replace report'" (click)="replaceReport(report.storageId)"></app-button>
-                  <app-button data-cy-test="compareReport" [title]="'Compare'" [text]="'Compare'" (click)="compareReports(report.storageId)"></app-button>
+                  <app-button data-cy-test="replaceReport" [text]="'Replace'" [title]="'Replace report'"
+                              (click)="replaceReport(report.storageId)"></app-button>
+                  <app-button data-cy-test="compareReport" [title]="'Compare'" [text]="'Compare'"
```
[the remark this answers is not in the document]
MatthijsSmets: See other comment.

src/app/test/test.component.html
```diff
@@ -5,79 +5,93 @@
...
+      <app-button data-cy-test="deselectAll" [title]="'Deselect All'" [text]="'Deselect All'"
+                  (click)="uncheckAll()"></app-button>
+      <app-button data-cy-test="deleteSelected" [icon]="'fa fa-trash'" [title]="'Delete Selected'"
+                  (click)="openDeleteModal()"></app-button>
       <app-button [text]="'Clone'" [title]="'Clone report'" (click)="openCloneModal()"></app-button>
-      <app-button data-cy-test="copySelected" [icon]="'fa fa-copy'" [title]="'Copy Selected'" (click)="copySelected()"></app-button>
-      <app-button data-cy-test="downloadBinary" [icon]="'fa fa-download'" [title]="'Download Binary'" (click)="downloadSelected()"></app-button>
+      <app-button data-cy-test="copySelected" [icon]="'fa fa-copy'" [title]="'Copy Selected'"
+                  (click)="copySelected()"></app-button>
+      <app-button data-cy-test="downloadBinary" [icon]="'fa fa-download'" [title]="'Download Binary'"
```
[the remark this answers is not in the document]
philipsens: Okay, I won't comment on it again

src/app/debug/table/table-settings-modal/table-settings-modal.component.html
MatthijsSmets: Fixed

src/app/test/test-folder-tree/test-folder-tree.component.ts
MatthijsSmets: Fixed

src/app/test/test.component.ts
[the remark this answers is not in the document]
philipsens: I've created a thingy for this in our conventions document. @Matthbo do you have any opinion on this?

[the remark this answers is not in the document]
Matthbo: I'd say using `+=` for "string building" is pretty clean and clear. If it were taken out of the current scope or put into a new scope (a string passed through to a different function) then using string interpolation is clearer and it returns a new string instead of doing a string manipulation.

src/app/test/test.component.ts
Matthbo: Interesting, seems like angular is parsing the template before the component is initialized according to their own lifecycle? That is kind of disappointing, but I can see why you would call this in the constructor now

src/app/test/test-folder-tree/test-folder-tree.component.ts
Matthbo: Shouldnt the try catch then cover the whole function body where the catch returns an empty array? Now you're not handling the error at all and if only one of the items has a wrong path it'll not be obvious that the error has actually occurred which then causes confusion maybe you but definitely anyone else working on this project

src/app/test/test.component.ts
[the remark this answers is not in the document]
MatthijsSmets: So you're saying it looks cleaner but consistency is more important and therefore the template is better in this case?

src/app/test/test-folder-tree/test-folder-tree.component.ts
Matthbo: You can return an empty array in the catch, also with the current code you don't even guarantee that you get an empty array back since other items can have valid paths and thus can the array be filled even with the error having been thrown

src/app/test/test.component.ts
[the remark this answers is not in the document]
Matthbo: No I'm saying that context should decide what the better option is, if you're making a string and manipulating that in the same function then using `+=` is a great solution. If you get a string passed through as an argument then using string interpolation to return a new string is a better solution instead. So yes using string interpolation here is a good solution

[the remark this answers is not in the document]
MatthijsSmets: I see your point, however on line 291 the string is prefixed with a / and on line 294 the string is suffixed with a /. Because consistency is key, i think its better to do both of them the same way. And because the += is not possible for prefixing, i think the template is the better option.

src/app/test/test.component.ts
```diff
@@ -246,25 +255,46 @@ export class TestComponent implements OnInit {
...
+    }
+  }
+
+  transformFilter(value: string): string {
+    if (value !== '' && value !== 'Reports') {
+      if (!value.startsWith('/')) {
+        value = `/${value}`;
+      }
+      if (!value.endsWith('/')) {
+        value = `${value}/`;
```
philipsens, line 294: Aren't these methods the same? The value assertion could be done in the calling function. Then these two functions can be merged.

src/app/test/test.component.ts
```diff
@@ -246,25 +255,46 @@ export class TestComponent implements OnInit {
...
+      const map: UpdatePathSettings = { path: path, action: action };
       this.httpService.updatePath(reportIds, this.currentView.storageName, map).subscribe(() => this.loadData(path));
     } else {
       this.toastService.showWarning('No Report Selected!');
     }
   }
 
+  addSlashesToPathEnds(path: string): string {
+    if (path.length > 0 && !path.startsWith('/')) {
+      path = `/ ${path}`;
```
philipsens, line 269: I think this space is an accident

src/app/test/test.component.ts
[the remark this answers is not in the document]
Matthbo: As I said, string interpolation (using template literals / template strings) is a good option because of the before mentioned reasons but I also like your reasoning so let's call this resolved for now

src/app/test/test.component.ts
MatthijsSmets: Fixed

src/app/test/test-folder-tree/test-folder-tree.component.ts
```diff
@@ -1,86 +1,100 @@
...
+    }
+    let folderName = path.slice(0, index);
+    const remainingPath = path.replace(folderName, '');
+    if (folderName.startsWith('/')) {
+      folderName = folderName.slice(1);
+    }
+    const treeItem: CreateTreeItem = { name: folderName, icon: 'assets/tree-icons/folder.svg' };
+    if (remainingPath && remainingPath !== '/' && remainingPath !== '') {
+      const child = this.createFolderFromPath(remainingPath);
+      treeItem.children = [child as Child];
```
Matthbo, line 89:
```suggestion
      const child: Child = this.createFolderFromPath(remainingPath);
      treeItem.children = [child];
```

philipsens (dismissed):

Matthbo (dismissed):

Matthbo (approved):

PR #401, opened 2024-05-10
Fix for the deleteAllTestReports command for Cypress tests

MarkvdVorst:
The if-statement in the deleteAllTestReports would always give a length of 1 and it was called asynchronously. Because it was asynchronous it caused a race condition somehow, making the test flacky.

Applied resetApp method where applicable now that the deleteAllTestReports works (which is called in the resetApp method)

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -40,7 +40,7 @@ Cypress.Commands.add('resetApp' as keyof Chainable, () => {
   );
   cy.deleteAllTestReports();
   cy.initializeApp();
-})
+});
 
 Cypress.Commands.add('navigateToTestTabAndWait' as keyof Chainable, () => {
   navigateToTabAndWait('test');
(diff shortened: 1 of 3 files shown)


cypress/support/commands.ts
```diff
@@ -205,16 +205,17 @@ Cypress.Commands.add('enableShowMultipleInDebugTree' as keyof Chainable, () => {
 //Clear reports in test tab if any present
 Cypress.Commands.add('deleteAllTestReports' as keyof Chainable, () => {
   cy.navigateToTestTabAndWait();
-  cy.get('[data-cy-test="selectAll"]').click();
-  cy.get('[data-cy-test="deleteSelected"]').click();
-  console.log(Cypress.$('[data-cy-delete-modal="confirm"]'));
-  if (Cypress.$('[data-cy-delete-modal="confirm"]').length > 0) {
-    cy.get('[data-cy-delete-modal="confirm"]').should('exist');
-  }
+  cy.get('[data-cy-test="table"]').should('exist');
```
MatthijsSmets, line 208: The assertion has to be removed if this is a cleanup function. It should not assert anything.

MatthijsSmets (changes requested):
Looks good, just a minor comment.

MatthijsSmets (approved):

PR #403, opened 2024-05-17
Error shown for wrong user input in filter

MarkvdVorst:
Should not be merged without merge of PR [#239](https://github.com/wearefrank/ladybug/pull/239) of the [ladybug](https://github.com/wearefrank/ladybug) project.

In the backend the types for each metadata is given through the API to the frontend. User input for the filter is now being validated. If it's the wrong type, it will display an error and will not submit the input given by the user to filter.

afbeelding

### DIFF
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
@@ -31,7 +31,6 @@ export class FilterSideDrawerComponent implements OnDestroy, OnInit {
 
   shouldShowFilterSubscriber!: Subscription;
   metadataLabelsSubscriber!: Subscription;
-  filterSubscriber!: Subscription;
   currentRecordsSubscriber!: Subscription;
   metadataTypesSubscriber!: Subscription;
   filterErrorsSubscriber!: Subscription;
@@ -63,16 +62,20 @@ export class FilterSideDrawerComponent implements OnDestroy, OnInit {
         this.metadataTypes = metadataTypes;
       },
     );
+    this.filterErrorsSubscriber = this.filterService.filterError$.subscribe(
+      (filterErrors: [boolean, Map<string, string>]): void => {
+        this.filters = filterErrors[1];
+      },
+    );
   }
 
   unsubscribeAll(): void {
     this.shouldShowFilterSubscriber.unsubscribe();
     this.metadataLabelsSubscriber.unsubscribe();
-    this.filterSubscriber.unsubscribe();
     this.currentRecordsSubscriber.unsubscribe();
   }
 
-  closeFilter() {
+  closeFilter(): void {
     this.filterService.setShowFilter(false);
   }
 
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
@@ -53,9 +53,6 @@ export class FilterSideDrawerComponent implements OnDestroy, OnInit {
     this.metadataLabelsSubscriber = this.filterService.metadataLabels$.subscribe((metadataLabels: string[]): void => {
       this.metadataLabels = metadataLabels;
     });
-    this.filterSubscriber = this.filterService.filterContext$.subscribe((filterContext: Map<string, string>): void => {
-      this.filters = filterContext;
-    });
     this.currentRecordsSubscriber = this.filterService.currentRecords$.subscribe(
       (records: Map<string, Array<string>>): void => {
         this.currentRecords = records;
--- a/src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
+++ b/src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
@@ -25,7 +25,6 @@ import { TitleCasePipe } from '@angular/common';
 export class FilterSideDrawerComponent implements OnDestroy, OnInit {
   protected shouldShowFilter!: boolean;
   protected metadataLabels!: string[];
-  protected filters: Map<string, string> = new Map<string, string>();
   protected currentRecords: Map<string, Array<string>> = new Map<string, Array<string>>();
   protected metadataTypes!: Map<string, string>;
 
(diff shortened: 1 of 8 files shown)


src/app/debug/table/table.component.ts
```diff
@@ -582,4 +589,31 @@ export class TableComponent implements OnInit, OnDestroy {
...
+          break;
+        }
+        case 'timestamp': {
+          typeLabel = 'date time';
+          break;
+        }
+      }
+      result += moreThanOne
+        ? ", Search value '" + value + "' is not a valid '" + typeLabel + "'"
+        : "Search value '" + value + "' is not a valid '" + typeLabel + "'";
```
Matthbo, line 614: Can't this be solved better with Template Literals?

MarkvdVorst: You're right. I added it in the new commit.

Matthbo (dismissed):

philipsens (changes requested):

src/app/debug/table/table.component.ts
```diff
@@ -582,4 +589,31 @@ export class TableComponent implements OnInit, OnDestroy {
...
+        }
+        case 'long': {
+          typeLabel = 'decimal number';
+          break;
+        }
+        case 'timestamp': {
+          typeLabel = 'date time';
+          break;
+        }
+      }
```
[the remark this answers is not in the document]
MarkvdVorst: I can add it, but string is currently always counted as correct input so it wouldn't ever be displayed. I'll add it in case I missed some scenario where it would give an error

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -24,14 +24,14 @@ import { TitleCasePipe } from '@angular/common';
...
+  protected metadataLabels!: string[];
   protected currentRecords: Map<string, Array<string>> = new Map<string, Array<string>>();
+  protected metadataTypes!: Map<string, string>;
 
   shouldShowFilterSubscriber!: Subscription;
-  metadataNamesSubscriber!: Subscription;
-  filterSubscriber!: Subscription;
+  metadataLabelsSubscriber!: Subscription;
   currentRecordsSubscriber!: Subscription;
+  metadataTypesSubscriber!: Subscription;
```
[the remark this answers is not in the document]
MarkvdVorst: If I don't remove the undefined from the subscription, a check for each subscription would have to be made for each subscription seperately to see if it's undefined, otherwise it will give an error.
Did you have something in mind for that? I think the same reason goes for why the other subscription remove undefined as well.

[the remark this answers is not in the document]
philipsens: Well you'd get: 
```typescript
unsubscribeAll(): void {
  this.shouldShowFilterSubscribtion?.unsubscribe();
  this.metadataLabelsSubscribtion?.unsubscribe();
  this.currentRecordsSubscribtion?.unsubscribe();
  this.metadataTypesSubscribtion?.unsubscribe();
}
```

But I think it is better than having an object that can be null after the constructor, but saying it isn't. It is also generally good to try to avoid it as much as possible, just so no wrong assumptions will go through.

Goed geëngelst.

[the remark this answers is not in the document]
MarkvdVorst: Alright. I added the change.

src/app/debug/filter-side-drawer/filter-side-drawer.component.ts
```diff
@@ -24,14 +24,14 @@ import { TitleCasePipe } from '@angular/common';
...
+  protected metadataTypes!: Map<string, string>;
 
-  shouldShowFilterSubscriber!: Subscription;
-  metadataNamesSubscriber!: Subscription;
-  filterSubscriber!: Subscription;
-  currentRecordsSubscriber!: Subscription;
+  shouldShowFilterSubscriber?: Subscription;
+  metadataLabelsSubscriber?: Subscription;
+  currentRecordsSubscriber?: Subscription;
+  metadataTypesSubscriber?: Subscription;
```
MatthijsSmets, line 34: Can you please call them subscription as that is exactly what they are, like the type suggests.

src/app/debug/filter-side-drawer/filter.service.ts
```diff
@@ -42,4 +70,31 @@ export class FilterService {
...
+  isInValidTimestamp(userInput: string, metadataName: string): boolean {
+    return this.metadataTypes.get(metadataName) == 'timestamp' && Number.isNaN(Date.parse(userInput));
+  }
+
+  isInValidNumber(userInput: string, metadataName: string): boolean {
+    const regex: RegExp = /^-?\d+(\.\d+)?$/;
+    return (
+      (this.metadataTypes.get(metadataName) == 'int' || this.metadataTypes.get(metadataName) == 'long') &&
+      (Number.isNaN(Number.parseFloat(userInput)) || !regex.test(userInput))
+    );
```
MatthijsSmets, line 98:
```suggestion
    const metadataType = this.metadataTypes.get(metadataName);
    return (
      (type == 'int' || type == 'long') &&
      (Number.isNaN(Number.parseFloat(userInput)) || !regex.test(userInput))
    );
```

src/app/debug/table/table.component.ts
```diff
@@ -569,7 +576,7 @@ export class TableComponent implements OnInit, OnDestroy {
 
   calculateViewDropDownWidth(): void {
     let longestViewName = '';
-    for (let [key, value] of Object.entries(this.viewSettings.views)) {
+    for (let [key] of Object.entries(this.viewSettings.views)) {
```
MatthijsSmets, line 579: Wouldn't this work as well?
```suggestion
    for (let key of this.viewSettings.views.keys()) {
```
Otherwise the following: 
```suggestion
    for (let key of Object.keys(this.viewSettings.views)) {
```

MatthijsSmets (dismissed):
Some minor suggestions

src/app/debug/filter-side-drawer/filter.service.ts
philipsens: I would like long `if` statements like these to be encapsulated.

MarkvdVorst: Changed it in the new commit

src/app/debug/table/table.component.ts
MarkvdVorst: went for the second option. first option did not work

philipsens (dismissed):

MatthijsSmets (approved):

PR #405, opened 2024-05-21
Open uploaded report in new tab

MarkvdVorst:
Instead of opening the uploaded file in the debug tab, it is now shown in its own seperate tab. This is mainly to give a better view of the uploaded report while not uploading it to the ladybug storage. This also removes some buttons that would not work with uploaded reports like the copy button.

Changed the Cypress test for the file upload.


src/app/report/report.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="report-tab row h-100">
   <div class="resizable-horizontal w-25 p-2 border-right h-100">
-    <jqxTree #treeReference (onItemClick)="selectReport($event)" [auto-create]="false"></jqxTree>
+    <jqxTree #treeReference (onItemClick)="selectReport($event)" [auto-create]="false" data-cy-debug="jqxTree"></jqxTree>
```
MatthijsSmets, line 3: Seeing as this tree is in the report component and not the debug, it should be:
```suggestion
    <jqxTree #treeReference (onItemClick)="selectReport($event)" [auto-create]="false" data-cy-report="jqxTree"></jqxTree>
```

cypress/e2e/debug_fileupload.cy.js
```diff
@@ -17,7 +14,6 @@ describe('Debug file upload', () => {
           fileName: 'testRerun.ttr',
         });
       });
-
-    cy.checkFileTreeLength(1);
+    cy.get('[data-cy-debug="jqxTree"]').find('li').should('have.length', 3);
```
MatthijsSmets, line 17:
```suggestion
    cy.get('[data-cy-report="jqxTree"]').find('li').should('have.length', 3);
```

MatthijsSmets (changes requested):

src/app/report/report.component.html
MarkvdVorst: added the change

cypress/e2e/debug_fileupload.cy.js
MarkvdVorst: added the change

MatthijsSmets (approved):

PR #417, opened 2024-05-30
Retyped multiple functions

Banshaan:
I changed the type and added changes to so some functions within test.component.ts that make use of the back-end functions and are exclusive to test.component.ts


src/app/test/test.component.ts
```diff
@@ -288,9 +293,9 @@ export class TestComponent implements OnInit, AfterViewInit {
   }
 
   updatePath(action: string): void {
-    let reportIds: string[] = this.helperService.getSelectedIds(this.reports);
+    const reportIds: string[] = this.helperService.getSelectedIds(this.reports);
     if (reportIds.length > 0) {
-      let path: string = this.moveToInputModel.value;
+      let path = (document.querySelector('#moveToInput')! as HTMLInputElement).value;
```
philipsens, line 298: Please use a `ViewChild` instead of a query selector.

src/app/test/test.component.ts
```diff
@@ -234,11 +238,12 @@ export class TestComponent implements OnInit, AfterViewInit {
     }
   }
 
-  uploadReport(event: any): void {
-    const file: File = event.target.files[0];
+  uploadReport(event: Event): void {
+    const eventTarget = event.target as HTMLInputElement;
+    const file: FileList | null = eventTarget.files;
```
philipsens, line 243: Please rename the variable too, it was a bit confusing at first because I saw `file[0]` first.
Why is the new code better than the old code?

src/app/test/test.component.ts
```diff
@@ -65,7 +67,7 @@ export class TestComponent implements OnInit, AfterViewInit {
   }
 
   ngAfterViewInit(): void {
-    this.loadData(null);
+    this.loadData('');
```
philipsens, line 70: Would an empty path mean the root of the project? Because than, I think it could be better to have an extra method called `loadRootData();` with passes the `this.testFileTreeComponent.rootFolder.name` to the `loadData`. It moves some of the logic out of the way, and then the check to see if the path is empty is more meaningful.

philipsens (changes requested):

src/app/test/test.component.ts
Matthbo: > Why is the new code better than the old code?

philipsens: Let me rephrase: What did you attempt to improve? I think it's less readable than what it used to be.

MatthijsSmets: I think if file is renamed to files than it is much more readable and now also typed, which was the goal of this PR.

Banshaan: Yeah my bad, it was supposed to be fileList. I forgot to change it.
What I was attempting to improve is that i gave the "Event" type to the input variable of the function, but that causes an error with `event.target` saying that it's possibly null and then when I put a question mark operator after `event.target` then it tells me that `files` doesn't exist on type `EventTarget`. That's how i got to this solution

philipsens: Then this would also be a solution. It would even check if the file exists, instead of just the array. 

```suggestion
    const file: File | undefined = eventTarget.files?.[0];
```

src/app/test/test.component.ts
Banshaan: I'm not sure i understand. Do you mean to call `this.loadRootData()` instead of `this.loadData()` and have `this.loadRootData()` give `loadData()` either the given path or the root path depending on if it has gotten an input?

philipsens: `loadRootData` shouldn't have any parameters and always give the root path to `loadData`. Then `loadData` doesn't need to worry about that anymore, cleaning it up somewhat.

src/app/test/test.component.ts
Banshaan: Yes, this seems to be working. I didn't realize the question mark operator was `?.`, I thought it was just the question mark.
Thanks for showing me that

src/app/test/test-folder-tree/test-folder-tree.component.ts
```diff
@@ -20,7 +20,7 @@ export class TestFolderTreeComponent {
     folderBehaviourOnClick: 'select',
     expandAllFolders: true,
     highlightOpenFolders: false,
-    autoSelectCondition: (item: CreateTreeItem) => this.autoSelectItem(item),
+    autoOpenCondition: (item: CreateTreeItem) => this.autoSelectItem(item),
```
MatthijsSmets, line 23: Why did you change this? autoOpenCondition is not the same as autoSelectCondition

MatthijsSmets, line 23: Why did you change this? autoOpenCondition is not the same as autoSelectCondition

MatthijsSmets (changes requested):

src/app/test/test.component.ts
philipsens: You're welcome!

src/app/test/test.component.ts
philipsens: Thank you for implementing it. Do you think you can also change every `loadData('')` to `loadRootData()`?

philipsens (changes requested):
See: https://github.com/wearefrank/ladybug-frontend/pull/417#discussion_r1622540497

src/app/test/test.component.ts
Banshaan: of course

src/app/test/test-folder-tree/test-folder-tree.component.ts
Banshaan: same as in the PR for retrieveRecords. I have fixed this mistake

src/app/test/test.component.ts
```diff
@@ -113,37 +115,43 @@ export class TestComponent implements OnInit, AfterViewInit {
...
         this.testFileTreeComponent.setData(this.reports);
         if (path) {
           setTimeout(() => {
             this.testFileTreeComponent.selectItem(path);
           });
         } else {
-          setTimeout(() => {
-            this.testFileTreeComponent.tree.selectItem(this.testFileTreeComponent.rootFolder.name);
-          });
+          this.loadRootData();
```
philipsens, line 142: Thank you for the change! But I wonder what would happen if `this.testFileTreeComponent.rootFolder.name` is empty. Would this get stuck in an infinite loop? In other cases, if the `path` is empty the method will be called again, having to subscribe (fetch data) again, being a bit wasteful, isn't it.
Maybe best to don't do anything (and trow an error) if a path is missing? (like what are you going to select then anyway :P). Also, `selectPath` might be a bit more descriptive than `path`.

Banshaan: I've looked into it and it does look like loadRootData is getting stuck in an infinite loop, but on the `ngOnInit` and the `ngAfterViewInit` already. I'm not sure I understand why this is happening, because when the code from `loadRootData` was still in `loadData` it was fine.

philipsens: Oh loadRootData only contains
```ts
loadRootData(): void {
    setTimeout(() => {
      this.testFileTreeComponent.tree.selectItem(this.testFileTreeComponent.rootFolder.name);
    });
  }
```
I thought it was
```ts
loadRootData(): void {
    loadData(this.testFileTreeComponent.rootFolder.name)
}
```
That would get stuck in a loop. But it is not like that, so never mind.

Banshaan: `this.testFileTreeComponent.rootFolder.name` is also selected differently than when you give `loadData` a `selectPath` input. 
For `loadData` the `selectItem` function is being called on `this.testFileTreeComponent` while for `this.testFileTreeComponent.rootFolder.name` the `selectItem` function is called on ``this.testFileTreeComponent.tree`.

Also with the current iteration of `loadRootData()` it is still getting stuck in an infinite loop. Is there anything I can do to fix this or should i put it back into the `loadData` function?

philipsens (approved):

Matthbo (approved):

MatthijsSmets:
If linter is fixed this can be merged

philipsens:
If e2e-test is fixed this can be merged

MatthijsSmets (approved):

PR #420, opened 2024-06-03
Refactor updatepath method in test component

### LINKED ISSUE #393: Refactor updatePath method in test.component.ts

MarkvdVorst:
Closes #393 

Replaced the updatepath calls in the button to setting a value in a FormControl variable and moved the updatePath() call to the ngSubmit of the form.

Found and fixed a small bug in the getFullPath() function where it would post two forward slashes. 
Ex. "//Simple report" instead of "/Simple report".


src/app/test/test.component.ts
```diff
@@ -48,6 +48,7 @@ export class TestComponent implements OnInit, AfterViewInit {
     storageName: 'Test',
     targetStorage: '',
   };
+  actionControl: FormControl<string> = new FormControl();
```
MatthijsSmets, line 51: Any reason you don't just have a variable for the value and a method to change that value? Why is it a form control in this case?

src/app/test/test.component.html
```diff
@@ -28,14 +28,13 @@
...
     <div class="form-group mt-2">
       <label for="moveToInput">Move/Copy to:</label>
-      <form class="input-group">
+      <form class="input-group" (ngSubmit)="updatePath()">
         <input #moveToInput type="text" class="form-control mr-1 my-2 py-0" id="moveToInput" name="filter"
                [ngModel]="currentFilter" />
-        <button type="submit" class="btn btn-info my-2 mx-1 px-3" (click)="updatePath('move')">Move</button>
-        <button type="submit" class="btn btn-info my-2 mx-1 px-3" (click)="updatePath('copy')"><i
+        <button type="submit" class="btn btn-info my-2 mx-1 px-3" (click)="actionControl.setValue('move')">Move</button>
+        <button type="submit" class="btn btn-info my-2 mx-1 px-3" (click)="actionControl.setValue('copy')"><i
```
MatthijsSmets, line 37:
```suggestion
        <button type="submit" class="btn btn-info my-2 mx-1 px-3" (click)="setUpdatePathAction('move')">Move</button>
        <button type="submit" class="btn btn-info my-2 mx-1 px-3" (click)="setUpdatePathAction('copy')"><i
```

src/app/test/test.component.ts
```diff
@@ -16,7 +16,7 @@ import { NodeLinkStrategy } from '../shared/enums/compare-method';
 import { TestFolderTreeComponent } from './test-folder-tree/test-folder-tree.component';
 import { ToastComponent } from '../shared/components/toast/toast.component';
 import { NgIf, NgFor } from '@angular/common';
-import { ReactiveFormsModule, FormsModule, NgModel } from '@angular/forms';
+import { ReactiveFormsModule, FormsModule, NgModel, FormControl } from '@angular/forms';
 import { ButtonComponent } from '../shared/components/button/button.component';
```
MatthijsSmets, line 21:
```suggestion
export const updatePathActionConst = ['move', 'copy'] as const;
export type UpdatePathAction = (typeof updatePathActionConst)[number]
```

src/app/test/test.component.ts
MatthijsSmets, line 51:
```suggestion
  updatePathAction: UpdatePathAction = 'move';
```

src/app/test/test.component.ts
```diff
@@ -287,12 +288,12 @@ export class TestComponent implements OnInit, AfterViewInit {
     this.reports.forEach((report) => (report.checked = false));
   }
```
MatthijsSmets, line 290:
```suggestion
setUpdatePathAction(value: UpdatePathAction) {
  this.updatePathAction = value;
}
```

src/app/test/test.component.ts
```diff
@@ -287,12 +288,12 @@ export class TestComponent implements OnInit, AfterViewInit {
...
 
-  updatePath(action: string): void {
+  updatePath(): void {
     let reportIds: string[] = this.helperService.getSelectedIds(this.reports);
     if (reportIds.length > 0) {
       let path: string = this.moveToInputModel.value;
       path = this.transformPath(path);
-      const map: UpdatePathSettings = { path: path, action: action };
+      const map: UpdatePathSettings = { path: path, action: this.actionControl.value };
```
MatthijsSmets, line 296:
```suggestion
      const map: UpdatePathSettings = { path: path, action: this.updatePathAction };
```

MatthijsSmets (changes requested):
I think its unnecessary to use a form control especially when not working with forms or inputs.

src/app/test/test.component.html
MarkvdVorst: added the change

src/app/test/test.component.ts
MarkvdVorst: I added the change. Should the type not be shown behind line 21 in the example for the sake of following conventions? I get an error when I try that however.

src/app/test/test.component.ts
MarkvdVorst: added the change

src/app/test/test.component.ts
MarkvdVorst: I added the change, but is there a reason for using a separate function to do this instead of assigning it directly in the html? Or is it just for typesafety since we do a check here for UpdatePathAction type?

src/app/test/test.component.ts
MarkvdVorst: added the change

src/app/test/test.component.ts
MarkvdVorst: I didn't know exactly what you meant in the issue by "Angular's built in form support" and this was what I came accross when doing some research on Angular forms. I see now that it's not that useful for this scenario.

src/app/test/test.component.ts
MatthijsSmets: I think its better to have the html just call functions

src/app/test/test.component.ts
MatthijsSmets: No it's fine like this, check the [conventions](https://github.com/frankframework/frontend-conventions/blob/master/typescript.md#enums)

MatthijsSmets (approved):

PR #422, opened 2024-06-10
refactor: try to refactor loading items and selecting path

philipsens:
@Banshaan Can you try if this code works. I don't have the development setup at my computer.

I've changed `loadData` a bit, if it breaks you can change that back.


MatthijsSmets (approved):

philipsens:
This was mend to make the code a bit cleaner, but the simple-tree gave an error, so this cant be used.

PR #473, opened 2024-07-04
Set version in an other way

philipsens:


MatthijsSmets (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@wearefrank/ladybug",
-  "version": "${project.version}",
+  "version": "[placeholder version]",
```
Matthbo, line 3: Wouldn't setting the version to a semver be better?
E.g. `0.0.0`?

philipsens: The value will be overwritten. Having an actual SemVer version can be confusing.

Matthbo: There might be a chance the package manager tries to show the version when running a cli command and it not being able to parse this string. Has this been tested for?

philipsens: So the POM first does a NPM version command, and then it's fixed.
Besides, it does show it, but because it is not being published or anything, it is not parsed.
Maybe in a multimodule project this will fail, but this is not. And if it were. Then this version should not even be needed...

I'm going to do a new PR in 3... 2... 1...

philipsens:
#475

PR #474, opened 2024-07-05
fix: debug tree indentation for end point checkpoints

### LINKED ISSUE #467: Indentation of nodes in debug tree is not what I expected
Indentation of nodes in debug tree is not what I expected. I tested this by running a test adapter that calls another adapter through a JavaListener.

image

image

This applies to all levels of the tree. In the user stories it is written that the checkpoint that closes a group is considered as the last child (story 1020). That would mean that the last child has to appear one position more to the right. And also that it has to disappear when the node is collapsed. If you collapse a node now, you keep the start and the end checkpoint.

This issue can also be fixed by updating story 1020.

MatthijsSmets:
Because the value for the type of checkpoint is sent as an integer starting at 1, I didn't see a way to use const string types as enums, that's why this solution goes againt our frontend conventions. I'm open to any suggestions

Closes #467.

### DIFF
--- a/src/app/debug/debug-tree/debug-tree.component.ts
+++ b/src/app/debug/debug-tree/debug-tree.component.ts
@@ -23,6 +23,7 @@ import {
 import { ButtonComponent } from '../../shared/components/button/button.component';
 import { NgIf } from '@angular/common';
 import { Checkpoint } from '../../shared/interfaces/checkpoint';
+import { CheckpointType } from '../../shared/enums/checkpoint-type';
 
 @Component({
   selector: 'app-debug-tree',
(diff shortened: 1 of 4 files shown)


src/app/debug/debug-tree/debug-tree.component.ts
```diff
@@ -155,35 +156,23 @@ export class DebugTreeComponent implements OnDestroy {
...
-        let currentStartpoint = startPointList;
+        const currentStartpoint: Checkpoint[] = startPointList;
         if (!currentStartpoint[startPointList.length - 1].checkpoints) {
           currentStartpoint[startPointList.length - 1].checkpoints = [];
         }
-        currentStartpoint[startPointList.length - 1].checkpoints!.push(checkpoints[i]);
-
-        if (checkpoints[i].type == 1) {
-          startPointList.push(checkpoints[i]);
+        currentStartpoint[startPointList.length - 1].checkpoints!.push(checkpoint);
```
philipsens, line 173: Hopefully this will be fixed in TS5.5!

src/app/shared/enums/checkpoint-type.ts
```diff
@@ -0,0 +1,23 @@
...
+  Startpoint = 1,
+  Endpoint,
+  Abortpoint,
+  Inputpoint,
+  Outputpoint,
+  Infopoint,
+  ThreadStartpointError,
+  ThreadStartpoint,
+  ThreadEndpoint,
+}
```
philipsens, line 11: We shouldn't use enums: https://github.com/frankframework/frontend-conventions/blob/master/typescript.md#enums

philipsens (changes requested):

src/app/shared/enums/checkpoint-type.ts
MatthijsSmets: I know, and I discussed this with @Matthbo but neither of us could figure out a good way that follows our conventions. Have a look at the way it's implemented and tell me how we should do it.

Matthbo: You could change the enum to use  `as const` instead but then you manually have to set the value for all the consecutive values. We discussed using an array instead so we could use the index as value but @MatthijsSmets didnt like the the `0` index being null/undefined.

philipsens: > but @MatthijsSmets didnt like the the 0 index being null/undefined.

Me neither. 

> but then you manually have to set the value for all the consecutive values.

That's unfortunate. I think it's a bit weird, tho, to have an exception to the rule. With all the type safety and such.

MatthijsSmets: I'm not a fan either, but the old solution was way worse.

Matthbo: I sent an hour last friday getting something better to work and it either got way too complex for this issue or I couldn't find a way to get it working properly with ts typings

philipsens: +1

philipsens (approved):

Issue #386, opened 2024-04-22
Replace current icons from file tree with svgs to reduce amount of assets


MatthijsSmets:
Currently there is a .gif file for every checkpoint type. A lot of files are the same icon with a different color, so this can be refactored to use svgs.

PR #499, opened 2024-07-12
Refactor: create class for transforming report to hierarchy structure

MatthijsSmets:
Created a class for transforming report to hierarchy structure for reusability.
Also added the getImage method here so the new class doesnt depend on the helper service.
The helper service still has the getImage method which will be removed when the jqx-tree is removed from the compare tab.

### DIFF
--- a/src/app/shared/classes/report-hierarchy-transformer.ts
+++ b/src/app/shared/classes/report-hierarchy-transformer.ts
@@ -1,0 +1,81 @@
+import { Report } from '../interfaces/report';
+import { Checkpoint } from '../interfaces/checkpoint';
+import { CHECKPOINT_TYPE_STRINGS, CheckpointType } from '../enums/checkpoint-type';
+
+export class ReportHierarchyTransformer {
+  private readonly THROWABLE_ENCODER: string = 'printStackTrace()';
+  private checkpointsTemplate: Checkpoint[] = [];
+  private startPointStack: Checkpoint[] = [];
+
+  transform(report: Report): Report {
+    const checkpoints: Checkpoint[] = report.checkpoints;
+
+    for (const checkpoint of checkpoints) {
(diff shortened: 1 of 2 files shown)


src/app/shared/classes/report-hierarchy-transformer.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    if (even) {
+      return `${img}-even.gif`;
+    }
+    return `${img}-odd.gif`;
+  }
+
+  private determineEvenCheckpoint(level: number): boolean {
+    return level % 2 == 0;
+  }
```
MarkvdVorst, line 80: To reduce the amount of code in the file, I think it would be better to remove the determineEvenCheckpoint method and just use conditional branching in its place as this function is only used once by the getImage() function.
```suggestion
  getImage(type: CheckpointType, encoding: string, level: number): string {
    let img = `assets/tree-icons/${CHECKPOINT_TYPE_STRINGS[type]}`;
    if (encoding === this.THROWABLE_ENCODER) {
      img += '-error';
    }
    return level % 2 == 0 ?
    `${img}-even.gif` :
    `${img}-odd.gif`
  }
```

MarkvdVorst (changes requested):
Looks good, I would change one small thing.

src/app/shared/classes/report-hierarchy-transformer.ts
MatthijsSmets: The function is there for readability, not reusability. I think using a ternary statement like this is not more readable.

MarkvdVorst: I don't think a whole function is needed for better readability in this case, but alright. It will be changed later for issue #386 anyway.

PR #506, opened 2024-07-15
Feat: report util

MatthijsSmets:
Utility class to check if report is of type report or of type checkpoint


MarkvdVorst (approved):
looks good

PR #507, opened 2024-07-15
Refactor/edit form component

ricklue:
Changed edit form component to form that is using FormGroup and FormControl instead of ngModel
Fixed discard functionality in difference modal
Refactored functionality to update a report in edit display component


src/app/report/edit-display/edit-display.component.ts
```diff
@@ -138,43 +149,28 @@ export class EditDisplayComponent {
...
-      this.getDifference('path', this.report.path, this.editFormComponent.path),
-      this.getDifference('transformation', this.report.transformation, this.editFormComponent.transformation),
-      this.getDifference('variables', this.report.variables, this.editFormComponent.variables),
+      this.editFormComponent.getDifference('name'),
+      this.editFormComponent.getDifference('description'),
+      this.editFormComponent.getDifference('path'),
+      this.editFormComponent.getDifference('transformation'),
+      this.editFormComponent.getDifference('variableCsv'),
     ];
   }
```
MatthijsSmets, line 158: Might be better to move all of this to the editFormComponent, that way the edit-display component doesn't need to know these string values

src/app/report/edit-form/edit-form.component.ts
```diff
@@ -1,53 +1,52 @@
...
+    this.editForm = new FormGroup(
+      {
+        name: new FormControl(this.report.name),
+        description: new FormControl(this.report.description),
+        path: new FormControl(this.report.path),
+        transformation: new FormControl(this.report.transformation),
+        variableCsv: new FormControl(this.report.variableCsv),
+      },
+      { updateOn: 'change' },
+    );
```
MatthijsSmets, line 30: Why is this in the ngOnChanges?

src/app/report/edit-form/edit-form.component.ts
```diff
@@ -1,53 +1,52 @@
...
-    if (this.report) {
-      this.initForm();
-    }
+    this.editForm = new FormGroup(
+      {
+        name: new FormControl(this.report.name),
+        description: new FormControl(this.report.description),
+        path: new FormControl(this.report.path),
+        transformation: new FormControl(this.report.transformation),
+        variableCsv: new FormControl(this.report.variableCsv),
```
MatthijsSmets, line 27: Create variables for these keys so you can reuse them, like we discussed

MatthijsSmets (changes requested):
Most of it looks good, just some small suggestions

MarkvdVorst (changes requested):

src/app/report/edit-form/edit-form.component.ts
ricklue: Do you mean why it is not in ngOnInit?

src/app/report/difference-modal/difference-modal.component.ts
```diff
@@ -56,6 +61,10 @@ export class DifferenceModalComponent {
   onClickSave(): void {
     if (this.saveOrDiscardType === 'save') {
       this.saveChangesEvent.next();
+    } else if (this.saveOrDiscardType === 'discard') {
+      this.discardChangesEvent.next();
+    } else {
+      this.toastService.showWarning('Something went wrong.');
```
MarkvdVorst, line 67: I would make it a little bit more specific. Maybe something like "Could not save/discard changes". This way we can more easily debug when something goes wrong and people using the frontend can get more information about what is going wrong.
What do you think @MatthijsSmets?

MarkvdVorst (changes requested):

src/app/report/difference-modal/difference-modal.component.ts
MarkvdVorst: true. then it's alright

MarkvdVorst (approved):

src/app/report/edit-form/edit-form.component.ts
MatthijsSmets: Yes

MarkvdVorst (dismissed):

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -84,9 +88,13 @@ export class EditDisplayComponent {
   showReport(report: Report): void {
     this.disableEditing();
     this.report = report;
-    report.xml
-      ? this.editor.setNewReport(report.xml)
-      : this.editor.setNewReport(this.helperService.convertMessage(report));
+    if (report.xml) {
+      this.editor.setNewReport(report.xml);
+      this.uid = undefined;
```
MatthijsSmets, line 93: I think it's better to use null here instead of undefined, undefined usually means that a variable was declared but does not have a value yet, while null is an assigned value.
Check [this post](https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript) for some examples.
```suggestion
      this.uid = null;
```

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -209,20 +195,28 @@ export class EditDisplayComponent {
...
-    const body = { stub: stubStrategy, ...this.getReportValues(checkpointId) };
+    const body = this.getReportValues(checkpointId);
 
     this.httpService.updateReport(storageId, body, this.currentView.storageName).subscribe({
       next: (response: any) => {
         response.report.xml = response.xml;
         this.report = response.report;
         this.saveReportEvent.next(this.report);
-        this.editor.setNewReport(this.report.xml);
+        this.editor.setNewReport(UpdateReportUtil.isUpdateCheckpoint(body) ? body.checkpointMessage : this.report.xml);
```
MatthijsSmets, line 205: Seeing as setNewReport only needs a string, couldn't you just use:
```suggestion
        this.editor.setNewReport(body.checkpointMessage ?? this.report.xml);
```

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -209,20 +195,28 @@ export class EditDisplayComponent {
...
     });
   }
 
+  discardChanges(): void {
+    this.disableEditing();
+    if (this.report.uid) {
+      this.editor.setNewReport(this.report.message);
+    }
+    this.toastService.showSuccess('Changes discarded!');
+  }
```
MatthijsSmets, line 218: So what happens if you're editing a report and not a checkpoint, don't you also need to reset the edit-form then?

MatthijsSmets (changes requested):
Looks good, just a few small things.

src/app/report/edit-display/edit-display.component.ts
ricklue: No, because body is from type UpdateReport | UpdateCheckpoint, so Typescript doesn't know if property checkpoinMessage exists without the UpdateReportUtil function

src/app/report/edit-display/edit-display.component.ts
ricklue: No, because the user goes back to read-only mode, so the form will reset automatically

src/app/report/edit-display/edit-display.component.ts
ricklue: I see now that I don't use that variable anymore, because I replaced it with a UpdateReportUtil function, so I removed the whole variable.

MatthijsSmets (approved):

MatthijsSmets (approved):

MarkvdVorst (approved):

PR #512, opened 2024-07-17
Refactor: refactor compare tab

MatthijsSmets:
Jqx-widgets package has been removed as well as jquery.
Compare tab has been refactored to use the ng-simple-file-tree package.


src/styles.css
```diff
@@ -138,3 +138,7 @@ html, body {
 .cursor-pointer {
   cursor: pointer;
 }
+
+.red {
+  color: red;
+}
```
MarkvdVorst, line 144: I think this is still here from when we were testing the ng-simple-file-tree. This should be removed.

MarkvdVorst (changes requested):
Looks good to me. You forgot to remove one thing.

src/styles.css
MatthijsSmets: No it's actually used in the compare-tree component

src/app/compare/compare-tree/compare-tree.component.ts
```diff
@@ -1,209 +1,156 @@
...
+  selectCheckpoint(tree: NgSimpleFileTree, item: FileTreeItem, checkpointToMatch: Checkpoint): void {
+    if (item.children) {
+      for (const child of item.children) {
+        const checkpoint = child.originalValue as Checkpoint;
+        if (this.getCheckpointId(checkpoint.uid) === this.getCheckpointId(checkpointToMatch.uid)) {
+          tree.selectItem(child.path);
+          return;
+        } else {
+          this.selectCheckpoint(tree, child, checkpointToMatch);
+        }
```
Matthbo, line 120: Why return on the end of the if and also have an else?

src/app/compare/compare-tree/compare-tree.component.ts
```diff
@@ -1,209 +1,156 @@
...
+          this.getCheckpointId(checkpoint.uid) === this.getCheckpointId(itemToMatch.uid)
+        ) {
+          return checkpoint;
+        }
+        if (checkpoint.checkpoints) {
+          return this.findCorrespondingCheckpoint(itemToMatch, checkpoint);
+        }
+      }
     }
+    return undefined;
```
Matthbo, line 150: I'm not sure I like undefined here, is there a reason it's undefined instead of null?
For context, I see undefined as something not defined in memory yet while null is when you can't return the actual object for some reason.
If this function is used to assign a variable (`const x`), having a null value would make more sense then having `const x` still returning undefined even after assignment

src/app/compare/compare.component.ts
```diff
@@ -41,60 +53,59 @@ export class CompareComponent implements OnInit, AfterViewInit {
...
 
-  getData(id: string): CompareData | undefined {
-    return this.tabService.activeCompareTabs.get(id);
+  private renderDiffs(leftSide: string, rightSide: string): void {
+    this.originalModel = Object.assign({}, this.originalModel, {
+      code: leftSide,
+    });
+    this.modifiedModel = Object.assign({}, this.originalModel, {
+      code: rightSide,
+    });
```
Matthbo, line 82: This can be changed to use the spread operator https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

Matthbo (changes requested):
The code looks good, I do have a few nitpicks

src/app/compare/compare-tree/compare-tree.component.ts
MatthijsSmets: Because it's in a for loop

src/app/compare/compare.component.ts
MatthijsSmets: You're right, this was some old code that just got moved, fixed it.

src/app/compare/compare-tree/compare-tree.component.ts
Matthbo: Ah my bad!

Matthbo (approved):

PR #517, opened 2024-07-18
Refactor: report component

MatthijsSmets:
Changed the way report component is styled based on feedback from @jacodg.
Also created custom directive for copying values on click and displaying tooltip with `Copied`


src/app/report/edit-display/edit-display.component.html
```diff
@@ -1,115 +1,90 @@
...
-            </select>
-          </div>
-        }
-      </div>
+      @if (report.stub) {
+        <div class="w-fit mx-1 my-0">
+          <select class="form-control" (change)="selectStubStrategy($event)">
+            <option [value]="-1" [selected]="report.stub === -1">Use report level stub strategy</option>
+            <option [value]="0" [selected]="report.stub === 0">Always stub this checkpoint</option>
+            <option [value]="1" [selected]="report.stub === 1">Never stub this checkpoint</option>
```
Matthbo, line 78: Wouldn't this be better with a ngModel?

MatthijsSmets: Yes you're right, I didn't change it in this PR because it's already quite large, I've made an issue for this on the refactoring board.

Matthbo (approved):

PR #520, opened 2024-07-19
Refactor: test component

MatthijsSmets:


src/app/test/test.component.ts
```diff
@@ -65,106 +66,94 @@ export class TestComponent implements OnInit, AfterViewInit {
...
-  }
-
-  getCopiedReports(): void {
-    this.httpService.getTestReports(this.currentView.metadataNames, this.currentView.storageName).subscribe({
-      next: (response: TestListItem[]) => this.addCopiedReports(response),
-      error: () => catchError(this.errorHandler.handleError()),
-    });
-  }
-
   loadData(path: string): void {
```
MarkvdVorst, line 97:
```suggestion
  loadData(path?: string): void {
```
Wouldn't it be better to have this as an optional paramater? Currently I only see one place where this method is called where it was not an empty string.

src/app/test/test.component.html
```diff
@@ -3,25 +3,37 @@
...
+      </button>
+      <button data-cy-test="copySelected" class="btn btn-info" title="Copy Selected"
+              (click)="copySelected()"><i class="fa fa-copy"></i></button>
+      <button data-cy-test="downloadBinary" class="btn btn-info" title="Download Binary"
+              (click)="downloadSelected()"><i class="fa fa-download"></i></button>
+      <button data-cy-test="upload" type="button" class="btn btn-info" title="Upload"
+              (click)="uploadFileTest.click()">
         <i class="fa fa-upload"></i>
         <input #uploadFileTest data-cy-test="uploadFile" type="file" id="uploadFileTest"
                (change)="uploadReport($event)">
```
MarkvdVorst, line 39: Is it necessary to have this extra input element? Wouldn't it be better to just the button or just the input element or am I missing something?

src/app/test/test.component.ts
MarkvdVorst: If we do this we can remove most of the falsy empty strings from the other method calls giving cleaner, more readable code.

MarkvdVorst (changes requested):
Looks good. I noted down some small points which may need some improvements if it's possible.

src/app/test/test.component.html
MatthijsSmets: Yeah the input element is there to allow native browser file uploads

MarkvdVorst: Alright. Good to know.

src/app/test/test.component.ts
MatthijsSmets: Oh right, good one, changed it.

MarkvdVorst (approved):

PR #528, opened 2024-07-22
Edit display typing

MarkvdVorst:
Part of refactoring the helper.service.ts

Added types to the edit display and helper.service.ts


src/app/shared/services/helper.service.ts
```diff
@@ -1,44 +1,42 @@
...
 
-  setButtonHtml(report: any, button: any, type: string, showConverted: boolean): void {
-    report.showConverted = showConverted;
-    button.target.title = `Convert to ${type}`;
-    button.target.innerHTML = type;
+  setButtonHtml(checkpoint: Checkpoint, button: MouseEvent, type: string, showConverted: boolean): void {
+    checkpoint.showConverted = showConverted;
+    const target: HTMLElement | null = button.target as HTMLElement;
+    target.title = `Convert to ${type}`;
+    target.innerHTML = type;
```
MatthijsSmets, line 39: This function should not be here anymore, this button with its functionality should be its own component

MarkvdVorst: I added the change. I also moved the convertMessage method fromthe helper service to the edit-display component as it was not used anywhere else.

src/app/report/edit-display/encoding-button/encoding-button.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  setButtonHtml(checkpoint: Checkpoint, button: MouseEvent, type: string, showConverted: boolean): void {
+    checkpoint.showConverted = showConverted;
+    const target: HTMLElement | null = button.target as HTMLElement;
+    target.title = `Convert to ${type}`;
+    target.innerHTML = type;
+  }
+
+  protected readonly ReportUtil = ReportUtil;
```
MatthijsSmets, line 50: This has to be refactored, just use a class variable for the text of the button

src/app/report/edit-display/encoding-button/encoding-button.component.html
```diff
@@ -0,0 +1,3 @@
+@if (ReportUtil.isCheckPoint(selectedNode) && selectedNode.encoding === 'Base64') {
```
MatthijsSmets, line 1: This if statement can be simplified to a class variable that is updated in `ngOnChanges`

MatthijsSmets (changes requested):

src/app/report/edit-display/encoding-button/encoding-button.component.ts
MarkvdVorst: added the change

src/app/report/edit-display/encoding-button/encoding-button.component.html
MarkvdVorst: added the change

src/app/report/edit-display/encoding-button/encoding-button.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  selector: 'app-encoding-button',
+  standalone: true,
+  imports: [],
+  templateUrl: './encoding-button.component.html',
+  styleUrl: './encoding-button.component.css',
+})
+export class EncodingButtonComponent implements OnChanges {
+  @Input({ required: true }) selectedNode!: ReportOrCheckpoint;
+  @Input({ required: true }) editor!: CustomEditorComponent;
+  buttonType: string = 'Base64';
```
MatthijsSmets, line 16: Does it always start at Base64?

src/app/report/edit-display/encoding-button/encoding-button.component.ts
```diff
@@ -0,0 +1,47 @@
...
+@Component({
+  selector: 'app-encoding-button',
+  standalone: true,
+  imports: [],
+  templateUrl: './encoding-button.component.html',
+  styleUrl: './encoding-button.component.css',
+})
+export class EncodingButtonComponent implements OnChanges {
+  @Input({ required: true }) selectedNode!: ReportOrCheckpoint;
+  @Input({ required: true }) editor!: CustomEditorComponent;
```
MatthijsSmets, line 15: This makes the flow of data sharing between components very unlogical IMO, right now this flow looks a little like:
```mermaid
flowchart TD
    A[edit-display.component] --> B[encoding-button.component];
    A <--> C[custom-editor.component]
    B --> C
```
however it should be:
```mermaid
flowchart TD
    A[edit-display.component] <--> B[encoding-button.component];
    A <--> C[custom-editor.component]
```
I think it makes more sense to have all data sharing go from the parent to the child, so in this case it's best to have the encoding-button.component have an output so the parent can do with that output as it chooses, this way our components have a tight cohesion but are loosely coupled.

MatthijsSmets (changes requested):

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -85,22 +90,33 @@ export class EditDisplayComponent {
...
+      this.editor.setNewReport(this.convertMessage(this.selectedNode));
+    }
     this.rerunResult = undefined;
     this.displayReport = true;
   }
 
-  changeEncoding(button: any): void {
-    this.editor.setNewReport(this.helperService.changeEncoding(this.report, button));
+  convertMessage(checkpoint: Checkpoint): string {
+    let message: string = checkpoint.message === null ? '' : checkpoint.message;
```
ricklue, line 106:
```suggestion
    let message: string = checkpoint.message ?? '';
```
I think this is a cleaner way to write this line of code, right?

src/app/report/edit-display/encoding-button/encoding-button.component.ts
MarkvdVorst: Yes, if the encoding is set to base64 it will always start as base64

src/app/report/edit-display/encoding-button/encoding-button.component.ts
MarkvdVorst: I moved the this.editor.setNewReport(message) call to a method in the edit-display component which gets called on te output attribute eventemitter from the encoding-button component.

src/app/report/edit-display/edit-display.component.ts
MarkvdVorst: Yes, I had not noticed that line yet. I added the change.

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -103,7 +103,7 @@ export class EditDisplayComponent {
   }
 
   convertMessage(checkpoint: Checkpoint): string {
-    let message: string = checkpoint.message === null ? '' : checkpoint.message;
+    let message: string = checkpoint.message ?? '';
```
MatthijsSmets, line 106: But if message is an empty string, this whole function does nothing right?

MarkvdVorst:
```suggestion
    let message: string = checkpoint.message;
```
I'll just change it to this for thread safety and leave it at that..

MarkvdVorst: Added the change.

MatthijsSmets (changes requested):

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -85,22 +90,33 @@ export class EditDisplayComponent {
...
+    let message: string = checkpoint.message;
+    if (checkpoint.encoding == 'Base64') {
+      message = btoa(message);
+    }
+    return message;
   }
 
   rerunReport(): void {
-    const reportId: number = this.report.storageId;
+    const node: ReportOrCheckpoint = this.selectedNode;
```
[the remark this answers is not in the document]
MarkvdVorst: I added this after you asked me to avoid a race condition/shared mutable state for the code. So I added a local variable so it avoids a shared mutable state in case selectedNode gets changed while the method is running.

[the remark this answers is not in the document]
MatthijsSmets: You're right, my bad.

src/app/report/edit-display/edit-display.component.html
```diff
@@ -66,16 +66,14 @@
           (click)="openDifferenceModal('save')"
         ><i class="fa fa-save"></i></button>
       }
-      @if (report.encoding === 'Base64') {
-        <button class="btn btn-info" title="Convert to Base64" (click)="changeEncoding($event)">Base64</button>
-      }
+      <app-encoding-button [selectedNode]="selectedNode" (updatedMessageEvent)="setNewReport($event)"></app-encoding-button>
 
-      @if (report.stub) {
+      @if (selectedNode && selectedNode.stub) {
```
MatthijsSmets, line 71: This should only be visible if report is a checkpoint

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -57,15 +59,17 @@ import { View } from '../../shared/interfaces/view';
...
 })
 export class EditDisplayComponent {
+  protected readonly ReportUtil = ReportUtil;
   @Input() id: string = '';
   @Input() containerHeight!: number;
   @Input({ required: true }) currentView!: View;
   @Input() newTab: boolean = true;
   @Output() saveReportEvent: Subject<any> = new Subject<any>();
-  @Output() closeReportEvent: Subject<void> = new Subject<void>();
+  @Output() closeReportEvent: Subject<Report> = new Subject<Report>();
```
MatthijsSmets, line 72: Why was this changed from void?
The selectedNode is passed by it's parent, so you don't have to pass which report is closed here as the parent already knows.

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -111,33 +126,59 @@ export class EditDisplayComponent {
...
-    this.saveChanges((event.target as HTMLOptionElement).value);
+    this.saveChanges(Number.parseInt((event.target as HTMLOptionElement).value));
   }
 
   openDifferenceModal(type: ChangesAction): void {
+    const node: Report | Checkpoint | undefined = this.selectedNode;
+    if (!node) {
+      this.toastService.showDanger('Could not find report to open difference modal');
+      return;
+    }
```
MatthijsSmets, line 174: I see these two lines used in multiple methods, maybe it's better to move this to a method that just returns a checkpoint or a report and shows a toast if it's null or undefined?

src/app/shared/components/display-table/metadata-table.component.ts
```diff
@@ -16,7 +16,7 @@ import { ReportUtil } from '../../util/report-util';
 export class MetadataTableComponent implements OnChanges {
   protected readonly ReportUtil = ReportUtil;
 
-  @Input({ required: true }) report!: Report | Checkpoint;
+  @Input({ required: true }) report?: Report | Checkpoint;
```
MatthijsSmets, line 19: If it's required, then it can't be undefined

MatthijsSmets (changes requested):

src/app/report/edit-display/edit-display.component.html
```diff
@@ -66,16 +66,14 @@
           (click)="openDifferenceModal('save')"
         ><i class="fa fa-save"></i></button>
       }
-      @if (report.encoding === 'Base64') {
-        <button class="btn btn-info" title="Convert to Base64" (click)="changeEncoding($event)">Base64</button>
-      }
+      <app-encoding-button [selectedNode]="selectedNode" (updatedMessageEvent)="setNewReport($event)"></app-encoding-button>
 
-      @if (report.stub) {
+      @if ((ReportUtil.isReport(selectedNode) || ReportUtil.isCheckPoint(selectedNode)) && selectedNode.stub) {
```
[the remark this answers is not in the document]
MatthijsSmets: My bad, this should only be type checkpoint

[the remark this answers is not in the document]
MarkvdVorst: Both report and checkpoint have it so just checking for undefined is good enough. Added the change

src/app/report/edit-display/edit-display.component.ts
MarkvdVorst:
```ts
closeReport(removeReportFromTree: boolean): void {
    this.displayReport = false;
    this.editingRootNode = false;
    this.editingChildNode = false;
    if (removeReportFromTree) {
      this.closeReportEvent.next(this.report);
    }
    this.editor.setNewReport('');
  }
  ```
  This was the code before the change. It gave the this.report which was of type any to the closeReportEvent. I don't know why angular allowed this to be a void, since a variable is being send anyway.
As it stands, the parent does not ask for the current report when removing but does have access to the current report. 

I'll make it so the closeReport method actually calls for the current report instead of asking for one from the event.

File: src/app/report/edit-display/edit-display.component.ts (hunk 11 above)
Comment (MarkvdVorst) [reply src/app/report/edit-display/edit-display.component.ts:174]: That's a good idea. I'll add the change.

File: src/app/report/edit-display/edit-display.component.html (hunk 13 above)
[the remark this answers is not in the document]
Comment (MarkvdVorst) [reply src/app/report/edit-display/edit-display.component.html]: Added the change

File: src/app/report/edit-display/edit-display.component.ts [hunk 14]
```diff
@@ -85,22 +90,33 @@ export class EditDisplayComponent {
     private errorHandler: ErrorHandling,
   ) {}
 
-  showReport(report: Report): void {
+  showReport(node: ReportOrCheckpoint): void {
```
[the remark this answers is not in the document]
MarkvdVorst: Added the change

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -111,33 +127,57 @@ export class EditDisplayComponent {
...
+    if (!node) {
+      this.toastService.showDanger('Could not find report to download');
+      return;
+    }
+    let queryString: string;
+    queryString = ReportUtil.isReport(node)
+      ? node.storageId.toString()
+      : ReportUtil.isCheckPoint(node)
+        ? node.uid.split('#')[0]
+        : '';
```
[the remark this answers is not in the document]
MarkvdVorst: Added the change

src/app/report/edit-display/edit-display.component.html
MarkvdVorst: Added the change.

src/app/report/edit-display/edit-display.component.ts
MarkvdVorst: Added the change.

src/app/report/edit-display/edit-display.component.ts
MarkvdVorst: Added the change

src/app/shared/components/display-table/metadata-table.component.ts
MarkvdVorst: Added the change

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -111,33 +127,57 @@ export class EditDisplayComponent {
...
     this.httpService.handleSuccess('Report Downloaded!');
   }
 
   selectStubStrategy(event: Event): void {
-    this.saveChanges((event.target as HTMLOptionElement).value);
+    this.saveChanges(Number.parseInt((event.target as HTMLOptionElement).value));
   }
 
   openDifferenceModal(type: ChangesAction): void {
+    const node: ReportOrCheckpoint = this.selectedNode;
```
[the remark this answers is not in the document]
MarkvdVorst: Added the change

MarkvdVorst:
Branch contains a flacky bug that shows up sometimes for the debugTreeGuardedCopyReport command from cypress tests

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -244,8 +285,9 @@ export class EditDisplayComponent {
     }
   }
 
-  toggleToolTip(toolTip: MatTooltip): void {
-    toolTip.show();
-    setTimeout(() => toolTip.hide(), 2500);
+  setNewReport(message: string) {
+    this.editor.setNewReport(message);
```
ricklue, line 289: You have made a setNewReport function, but you don't use it? Or is this only for the encoding button?

src/app/report/edit-display/edit-display.component.ts
```diff
@@ -293,20 +285,9 @@ export class EditDisplayComponent {
...
   }
 
-  getSelectedNode(): Report | Checkpoint | undefined {
-    const node: Report | Checkpoint | undefined = this.selectedNode;
-    if (!node) {
-      this.toastService.showDanger('Selected report was undefined');
-    }
-    return node;
-  }
+  protected readonly Number: NumberConstructor = Number;
```
MatthijsSmets, line 292: Please move this to the top of the file just like ReportUtil

MatthijsSmets (dismissed):

src/app/shared/components/display-table/metadata-table.component.ts
MatthijsSmets: Yes it should even

src/app/report/edit-display/edit-display.component.ts
MarkvdVorst: Added the change.

ricklue (approved):

PR #530, opened 2024-07-24
Fix: debug tree does not fit

### LINKED ISSUE #484: Debug tree past niet
De inhoud van de debug tree past niet in het paneel. Ik kan het eind niet te pakken krijgen en zie ook geen scroll bar om het tevoorschijn te halen.

image

Dit treedt op in frontend versie 0.1.0-20240710.153412.

MatthijsSmets:
Fix debug tree styling so scrolling works normally, also the button bar is now sticky and doesn't jitter when scrolling.
Closes #484


ricklue (approved):
Looks good and it works fine

PR #531, opened 2024-07-24
Fix: open same report

### LINKED ISSUE #37: The same report can be opened multiple times
This issue was found using the ladybug-frontend project combined with the ibis-ladybug-test-webapp example project. With "ng serve" in the ladybug-frontend project, the Angular GUI is served. It calls the backend implemented in projects ibis-ladybug and ibis-ladybug-test-webapp. The ibis-ladybug-test-webapp also offers a .jsp page that can be used to create reports. With this setup, the issue can be reproduced as follows:
* Create two reports using the .jsp page.
* Open Ladybug.
* Click a report in the table. It is opened.
* Click "Open All". Both reports are opened. The report that was opened in the previous step appears twice now in the view of open reports.

### LINKED ISSUE #478: Toggling Show multiple files in debug file-tree setting has some weird behavior
I enable the `Show multiple files in debug file-tree` setting and open a bunch of files in the tree.\
image

After that I close them all
image

Then if I disable the `Show multiple files in debug file-tree` setting again, it opens the last closed report.
image

MatthijsSmets:
Clicking a report that's already openen in the debug-tree will now select it, instead of opening again.
If multiple reports are opened, clicking a node doesnt also select other nodes with the same name anymore.
Closes #37.
Closes #478.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "eslint-plugin-unicorn": "^54.0.0",
     "font-awesome": "^4.7.0",
     "monaco-editor": "0.44.0",
-    "ng-simple-file-tree": "^0.1.19",
+    "ng-simple-file-tree": "^0.1.20",
     "ngx-monaco-editor-v2": "17.0.1",
     "path": "^0.12.7",
     "popper.js": "^1.16.1",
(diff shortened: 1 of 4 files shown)


ricklue (approved):
Looks good

PR #535, opened 2024-07-24
Refactor: add possibly null

MatthijsSmets:
Added possibly null types to some properties of `Report` and `Checkpoint`.
Refactored meta-data table component to use types.
Removed redundant pipe.


src/app/shared/components/display-table/metadata-table.component.html
```diff
@@ -99,10 +100,6 @@
         <div class="key">Message class name</div>
         <div class="value" [appCopyTooltip]="report.messageClassName">{{ report.messageClassName }}</div>
       </div>
-      <div class="table-col">
-        <div class="key">Path</div>
-        <div class="value" [appCopyTooltip]="report.path">{{ report.path }}</div>
-      </div>
```
ricklue, line 105: Does it have a special reason why this column is removed?

MatthijsSmets: yes, because neither reports or checkpoints have the `path` property

ricklue (approved):

PR #543, opened 2024-07-26
fix: rerunning report

### LINKED ISSUE #537: Rerunning a report does not work - toast that generator is not enabled
I copied a report to the Test tab and pressed the rerun button. I got a toast saying that the generator is not enabled. But when I went back to the Debug tab and opened the settings dialog, I saw that the report generator was enabled.

image

I tested with the frontend obtained from `ng serve` and ladybug-frontend commit 152d2aada1ba2c09d76cf5be5139e825ad836208. I had ladybug (backend) commit 01654ba507758b9a9fd3348598dfb9e9d4f40b2f.

MatthijsSmets:
Rerunning a report in the test tab is now fixed, however replacing the report does not work yet, and will be implemented later.
Closes #537.


src/app/test/test.component.html
```diff
@@ -108,21 +115,20 @@
...
                     </span>
                   }
                 </td>
                 <td>
-                  @if (getReranReport(report.storageId)) {
+                  @if (report.reranReport) {
                     <span class="buttons">
                         <button data-cy-test="replaceReport" class="btn btn-info" title="Replace report"
-                                (click)="replaceReport(report.storageId)">Replace</button>
+                                disabled (click)="replaceReport(report)">Replace</button>
```
ricklue, line 129: Is this disabled because the functionality to replace a report is not implemented yet?

MatthijsSmets: Yes, I added disabled to the button and added a toast warning to the method that is called by the button to make it extra clear

ricklue: Do we have an issue for that already? Maybe in the backend?

MatthijsSmets: Yes #544

ricklue (approved):

PR #564, opened 2024-07-31
Refactor: stub strategy

ricklue:
Refactored stub strategy dropdown that it is using ngModel
When changing the stub strategy, this will open the difference modal, so the user have a confirmation about the change


MarkvdVorst (dismissed):

src/app/report/edit-display/edit-display.component.ts
MatthijsSmets, line 1: Previously changing the stub strategy triggered the `saveChanges` method, but now it doesn't, shouldn't you also add an `ngModelChange` as well to trigger the `saveChanges` method?

ricklue: We had a discussion if changing the stub strategy should trigger the `saveChanges` method or should trigger a function where the save button appears, so that was still a question that has to be answered.

MarkvdVorst (changes requested):
image
When changing the stub strategy, the checkpoint is seen as a report node and then changes the stub strategy dropdown to a rerun button. I don't think this should happen

MarkvdVorst (changes requested):
There are still two more bugs that I would like to see fixed in this PR if possible.

1. The opened report in the debug view does not update when you save edits.

steps to reproduce:
  - edit report
  - save report
  - go to different checkpoint
  - go to original checkpoint that was edited

The changes do no occur anymore when I switch. Closing and opening the report does show the correct change however.

2. When I edit a checkpoint, add some changes to it and then change the stub strategy, the changes get discarded.
steps to reproduce are the same as above.

The difference between this bug and the previous one is that even with closing and opening the report, the changes did not get saved and are not shown.

src/app/report/difference-modal/difference-modal.component.html
```diff
@@ -20,11 +20,17 @@ <h3 class="modal-title">Are you sure you want to {{ saveOrDiscardType }} changes
...
-              <div class="changeBox text-left p-1 overflow-auto">
-                @for (chunk of part.difference; track chunk[1]) {
-                  <span [style.background-color]="getChunkColorForDifferenceModal(+chunk[0])">{{ chunk[1] }}</span>
-                }
-              </div>
+              @if (stubChange) {
+                <div class="changeBox text-left p-1 overflow-auto"
+                >{{ part.difference }}
+                </div>
+              } @else if (!stubChange) {
```
MarkvdVorst, line 27: This should be just an else

MarkvdVorst (changes requested):
Looking better. I got a new bug now however. I noticed that you made it so the stub strategy cannot be changed while changes are not saved, however when I save the changes inside the edited checkpoint I still get the same warning message. Only switching checkpoints or closing and opening the report fixes this.

image

MarkvdVorst (approved):
Looks good. I notice in a lot of cases where multiple toasts are outputted. but for the sake of this issue it's not important for this PR. I'll make a new issue for it on the refactor board and we'll wait to see what @MatthijsSmets thinks about it.

see issue #610

src/app/report/difference-modal/difference-modal.component.html
ricklue: Added the change

PR #565, opened 2024-07-31
Wildcard functionality for the table filter.

### LINKED ISSUE #560: Filtering doesn't work with wildcards anymore like the tooltip suggests
When hovering a filter input field it tells the user multiple ways it can filter values:
`Search case insensitive using * as the wildcard character. Wildcards are automatically added at the beginning and the end unless the search value starts with [ and ends with ] or already starts or ends with the wildcard. The search is done case sensitive when the search value starts with [[ and ends with ]]. A regular expression search is done when the search value starts with ( and ends with ). In regular expression mode the null object is the same as the empty string. When the search value consists of the 4 characters null it will match when the object searched for is null. Use [] to match the empty string and [null] to match the string of the 4 characters null.`

This does not work however, entering a wildcard for storageId gives the error: 

Image

While we are at it, move those error messages to the input field that generates them, the current location above the table is not logical.

Also, clicking the clear button next to the input field does not remove the filter pill shown above the table, it only clears the value in the input field.

MarkvdVorst:
Closes #560 

Also added new test for the wildcard functionality and renamed some tests to follow conventions.


cypress/e2e/debug_filterTable.cy.js
```diff
@@ -7,11 +7,9 @@ describe('Table size and toggle filter', () => {
     cy.initializeApp();
   });
 
-  afterEach(() => {
-    cy.resetApp();
-  });
+  afterEach(() => cy.resetApp());
 
-  it('Typing in a table size and retyping it', () => {
+  it('Should change table size when changing display amount', () => {
```
MatthijsSmets, line 12: Yes thank you 🙌

src/app/debug/filter-side-drawer/filter.service.ts
```diff
@@ -30,7 +30,11 @@ export class FilterService {
       let errorFound: boolean = false;
       this.filterErrors.clear();
       for (let [key, value] of context) {
-        if (this.isInValidTimestamp(value, key) || this.isInValidNumber(value, key)) {
+        const metadataType: string | undefined = this.metadataTypes.get(key);
+        if (
+          (metadataType == 'timestamp' && !this.isValidTimestamp(value)) ||
+          ((metadataType == 'int' || metadataType == 'long') && !this.isValidNumber(value))
```
MatthijsSmets, line 36: Maybe move this assertion to it's own method, I have a hard time reading this.

MatthijsSmets (changes requested):

ricklue (approved):

PR #567, opened 2024-07-31
Fix: Report run error specification

### LINKED ISSUE #429: Unable to correlate report run error to specific report
In old GUI:
image

In news GUI the generic error toaster is used:
image

When running 1 or a few report this is not such a problem but when you have a 100 reports and some of the them fail it will be hard to know which error belongs to which report

ricklue:
Closes #429 
Error message is now shown in the test table instead of a toast, so it is clear which error correlates for the specific report.

### DIFF
--- a/src/app/test/test.component.ts
+++ b/src/app/test/test.component.ts
@@ -4,7 +4,7 @@ import { CloneModalComponent } from './clone-modal/clone-modal.component';
 import { TestSettingsModalComponent } from './test-settings-modal/test-settings-modal.component';
 import { TestResult } from '../shared/interfaces/test-result';
 import { ReranReport } from '../shared/interfaces/reran-report';
-import { catchError } from 'rxjs';
+import { catchError, Observable, of } from 'rxjs';
 import { Report } from '../shared/interfaces/report';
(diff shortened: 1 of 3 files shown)


src/app/test/test.component.ts
```diff
@@ -153,7 +154,11 @@ export class TestComponent implements OnInit {
         next: (response: TestResult): void => {
           report.reranReport = this.createReranReport(response);
         },
-        error: () => catchError(this.errorHandler.handleError()),
+        error: () =>
+          catchError((error: HttpErrorResponse): Observable<any> => {
+            report.error = error.message;
+            return of(error);
+          }),
```
MatthijsSmets, line 161: Just wondering, if the http call returns a 500 error from nginx for example, will it paste the entire 500 error html page in the table?

ricklue: In the error-handling service we are already using 'error.message', but I was not able to test it, so I am not 100% sure...

MatthijsSmets: We'll leave it like this for now, but eventually that should be refactored as well.

PR #571, opened 2024-08-01
refactor: test component

### LINKED ISSUE #488: If you press "Compare" for a failed report, the test result is gone in the Test tab
image

You see here a failed test report. If you press "Compare", a compare tab opens as expected. But if you return to the test tab, the red text about the failure is gone. The Compare and Replace buttons are also gone - it is as if the report was never executed.

MatthijsSmets:
Test tab now has a service that keeps track of all it's reports, that way switching tabs keeps the run-results in the table.
Refactored some css classes.
Toast component was set in multiple template files which is redundant, removed all but the one in app.component.
Closes #488.


src/app/test/test.component.html
```diff
@@ -9,33 +9,15 @@
...
+      <button class="btn btn-info" title="Options" (click)="testSettingsModal.open()"><i class="fa fa-cog"></i></button>
+      <button data-cy-test="deleteSelected" class="btn btn-info" title="Delete Selected" (click)="openDeleteModal()"><i class="fa fa-trash"></i></button>
+      <button class="btn btn-info" title="Clone report" (click)="openCloneModal()">Clone</button>
+      <button data-cy-test="copySelected" class="btn btn-info" title="Copy Selected" (click)="copySelected()"><i class="fa fa-copy"></i></button>
+      <button data-cy-test="downloadBinary" class="btn btn-info" title="Download Binary" (click)="downloadSelected()"><i class="fa fa-download"></i></button>
       <button data-cy-test="upload" type="button" class="btn btn-info" title="Upload"
               (click)="uploadFileTest.click()">
         <i class="fa fa-upload"></i>
-        <input #uploadFileTest data-cy-test="uploadFile" type="file" id="uploadFileTest"
+        <input #uploadFileTest data-cy-test="uploadFile" type="file" [hidden]="true"
```
ricklue, line 20: Why is this input always hidden?

cypress/support/commands.ts
```diff
@@ -77,6 +77,7 @@ Cypress.Commands.add('createReport' as keyof Chainable, (): void => {
   // No cy.visit because then the API call can happen multiple times.
   cy.request(`${Cypress.env('backendServer')}/index.jsp?createReport=Simple%20report`).then((resp: Cypress.Response<ApiResponse>) => {
     expect(resp.status).equal(200);
+    console.log(resp);
```
ricklue, line 80: Forgot to remove console log

ricklue (commented):
Two comments

src/app/test/test.component.html
MatthijsSmets: It's for uploading files, but the default input element cannot be styled very nicely.

ricklue (approved):

PR #572, opened 2024-08-01
fix: remove extra scrollbar

MatthijsSmets:
removed extra scroll bar from debug tab


mhdirkse:
There is no-one to properly review this code and @MatthijsSmets wants to merge. It is a small fix so I can live with this, provided that Mark and Rick are notified.

mhdirkse (approved):

PR #596, opened 2024-08-07
Refactor of test component

### LINKED ISSUE #539: Refactor matches method in test.component
This method is used in the template to determine is a report should be shown in the table in the test component.
```ts
  matches(report: TestListItem): boolean {
    const name = report.path + report.name;
    return new RegExp(`(/)?${this.currentFilter}.*`).test(name);
  }
```
There's a few problems with it however.

1. It's a method call in the template which means it's called everytime angular runs it's change detection, which is an awful lot. I think using a sorted list that gets updated whenever the filter is updated is a much more optimized way.
3. The path property is used to match a report, however if the report is in the root folder: `Reports` it's path is null.  Which means the `matches` method checks null/nameOfReport which is prone to errors. -> example:
```json
{
  "storageId": "1",
  "name": "Another simple report",
  "path": null,
  "description": null,
  "variables": null,
  "checked": true
}
```
However, and here is where it get's weird, if I copy a report to test tab it has that null value as path, but if I move it to a folder using the `Move/Copy to` input field and then move it back to root, it has a value of '/':
```json
{
  "storageId": "1",
  "name": "Another simple report",
  "path": "/",
  "description": null,
  "variables": null,
  "checked": true
}
```

MarkvdVorst:
Closes #539 

- Changed names in testtab.cy.ts to more conventional names.
- Moved table from test tab to new component.
- Refactored matches method so it only gets called when report data is updated.
- Refactored matches method so report.path can no longer be null. It now will default to an empty string. This way the value is still falsy and will avoid some errors that might appear in other places in the program.


src/app/test/test.component.html
```diff
@@ -35,98 +35,13 @@
...
-                    <span data-cy-test="runResult"
-                          [style.color]="report.reranReport.color">
-                      {{ report.reranReport.resultString }}
-                    </span>
-                  }
-                  @if (report.error) {
-                    <span style="color: red">
-                      {{ report.error }}
-                    </span>
-                  }
```
ricklue, line 109: In the new test table component, you haven't add this to the table

MarkvdVorst: Added it back again. Also noticed there was a missing disabled attribute for the replaceReport button which is now also added.

src/app/test/test.component.ts
```diff
@@ -299,39 +280,27 @@ export class TestComponent implements OnInit {
...
+    if (reportIds.length > 0) {
+      const path: string = this.transformPath(this.moveToInputModel.value);
+      const map: UpdatePathSettings = { path: path, action: this.updatePathAction };
+      this.httpService.updatePath(reportIds, this.testReportsService.storageName, map).subscribe({
+        next: () => this.loadData(path),
+        error: () => catchError(this.errorHandler.handleError()),
+      });
+      this.matches();
+      this.testReportsService.getReports();
+      this.refresh();
```
ricklue, line 301: This methods are called before the request is done. Is that correct?

MarkvdVorst: you're right. I didn't think of that yet. I added the change.

src/app/test/test-table/test-table.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    const amountSelectedSubscription: Subscription = this.testReportsService.amountSelected$.subscribe({
+      next: (value: number): void => {
+        this.amountOfSelectedReports = value;
+      },
+      error: () => catchError(this.errorHandler.handleError()),
+    });
+    this.subscriptions.add(amountSelectedSubscription);
+  }
+
+  unsubscribeToSubscriptions(): void {
```
MatthijsSmets, line 85: I know you moved this, but the incorrect English is bothering me 😁
```suggestion
  unsubscribeFromSubscriptions(): void {
```

src/app/test/test-table/test-table.component.ts
```diff
@@ -0,0 +1,160 @@
...
+      error: () => catchError(this.errorHandler.handleError()),
+    });
+  }
+
+  getFullPaths(): void {
+    for (const report of this.reports) {
+      report.fullPath = '';
+      report.fullPath = report.path
+        ? `${report.path.replace(this.currentFilter, '')}${report.name}`
+        : `/${report.name}`;
```
MatthijsSmets, line 110: Why is `fullPath` assigned to an empty string first if it's always reassigned after?

src/app/test/test-table/test-table.component.ts
```diff
@@ -0,0 +1,160 @@
...
+  imports: [FormsModule],
+  templateUrl: './test-table.component.html',
+  styleUrl: './test-table.component.css',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TestTableComponent implements OnChanges, OnInit, OnDestroy {
+  @Input() reports: TestListItem[] = new Array<TestListItem>();
+  @Input() currentFilter: string = '';
+  @Input() showStorageIds?: boolean;
+  @Output() toggleCheckEvent: EventEmitter<TestListItem> = new EventEmitter<TestListItem>();
```
MatthijsSmets, line 37: I believe checking and unchecking reports can just be handled from this table component right?

src/app/test/test-table/test-table.component.ts
```diff
@@ -0,0 +1,160 @@
...
+  @Output() runEvent: EventEmitter<TestListItem> = new EventEmitter<TestListItem>();
```
MatthijsSmets, line 38: Rerunning a report can also be handled from this component right?

MatthijsSmets (changes requested):
I have some questions

src/app/test/test-table/test-table.component.ts
```diff
@@ -0,0 +1,160 @@
...
+import { ToastService } from '../../shared/services/toast.service';
+import { FormsModule } from '@angular/forms';
+
+@Component({
+  selector: 'app-test-table',
+  standalone: true,
+  imports: [FormsModule],
+  templateUrl: './test-table.component.html',
+  styleUrl: './test-table.component.css',
+  changeDetection: ChangeDetectionStrategy.OnPush,
```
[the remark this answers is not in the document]
MarkvdVorst: Angular added that, but yes

src/app/test/test-table/test-table.component.ts
MarkvdVorst: The test component calls the run method from the runSelected method, so the method cannot just be moved to the test-table.
I think it will be better if we keep it like this.

src/app/test/test-table/test-table.component.ts
```diff
@@ -0,0 +1,160 @@
...
+})
+export class TestTableComponent implements OnChanges, OnInit, OnDestroy {
+  @Input() reports: TestListItem[] = new Array<TestListItem>();
```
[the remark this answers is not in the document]
MarkvdVorst: I first made it a required true, but for some reason the karma tests disagree with it being assigned and say that reports is undefined. I'll change it to the [] instead, but with the code now it cannot be required true

src/app/test/test-reports.service.ts
```diff
@@ -7,18 +7,20 @@ import { Observable, ReplaySubject, Subject } from 'rxjs';
   providedIn: 'root',
 })
 export class TestReportsService {
+  private testReportsSubject: Subject<TestListItem[]> = new ReplaySubject<TestListItem[]>(1);
+  private amountSelectedSubject: Subject<number> = new Subject<number>();
   metadataNames: string[] = ['storageId', 'name', 'path', 'description', 'variables'];
   storageName: string = 'Test';
-  private testReportsSubject: Subject<TestListItem[]> = new ReplaySubject<TestListItem[]>(1);
   testReports$: Observable<TestListItem[]> = this.testReportsSubject.asObservable();
+  amountSelected$: Observable<number> = this.amountSelectedSubject.asObservable();
```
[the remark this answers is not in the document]
MarkvdVorst: Added the change

src/app/test/test-table/test-table.component.ts
MarkvdVorst: hahaha, don't worry. At other places I used From instead of To. Added the change.

src/app/test/test-table/test-table.component.ts
MarkvdVorst: Added the change

src/app/test/test-table/test-table.component.ts
MarkvdVorst: You're correct. I moved the toggleCheck method to test table component

src/app/test/test-table/test-table.component.ts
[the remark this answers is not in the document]
MatthijsSmets: Why can't it be required?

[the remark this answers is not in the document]
MarkvdVorst: Added the change like we discussed

src/app/test/test-table/test-table.component.html
MatthijsSmets, line 1: the table is not scrollable

MatthijsSmets (changes requested):

MatthijsSmets (approved):

PR #604, opened 2024-08-12
Compare tab window using full height

### LINKED ISSUE #585: Diff in Compare tab doesn't take the space available at the bottom of the window
image

MarkvdVorst:
Closes #585 
After doing some research, here's my conclusion on the matter.
The monaco diff editor cannot be adjusted in height and should not. It always needs a set height. Therefore I made it so the entire remaining height that the element can use is filled up by the element.

For a better look, I added the following line to the monaco editor options:
`scrollBeyondLastLine: false`

Could not find a way to remove the big scrollbar on the right. This should be researched further at a later time.

### DIFF
--- a/src/app/compare/compare.component.css
+++ b/src/app/compare/compare.component.css
@@ -26,3 +26,7 @@
   padding-top: 1.25rem;
   height: 100%;
 }
+
+.diff-editor {
+  height: 100%;
+}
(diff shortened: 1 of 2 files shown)


ricklue (approved):
I think we should make an new issue for the big scrollbar, but for now it looks good

MarkvdVorst:
> I think we should make an new issue for the big scrollbar, but for now it looks good

I created the issue for it. See issue #611

PR #607, opened 2024-08-14
Filter transform when view box changes

### LINKED ISSUE #480: Filter options should not reset when you change views
Please do the following test.
* Make sure there are some reports in the report table in the debug tab.
* Apply some filters to the columns of this table.
* Select another view.
* The filters should not change and they should not be removed.

### LINKED ISSUE #542: If you switch between views that have different columns for the report table, you can jeopardize filtering
I tested this by combining ladybug with the FF! and by starting it using the Frank!Runner. I tweaked the FF! by adding an extra view. The new view is like the white box view but it has less columns for the report table. I did some manipulations to change the filter and to switch views. I managed to see a report that does not satisfy the filters, see below:

image

This example is not only of theoretical importance. Integration specialists do add columns to ladybug in practice, for example in ZaakBrug.

MarkvdVorst:
Closes #480 
Closes #542 

- In table.component.ts, the assigning for filters is now done by creating a new map with the filters in it. This is done so the active-filter component is properly updated. 
- In table.component.ts the metadataLabels were passed to the filterService but that should be metadataNames. 
- Filters are now transformed to new state if view box is changed where some columns do not exists. It deletes the values from the map. 
- Added new cypress test for previous statement.


cypress/e2e/debug/filterTable.cy.ts
```diff
@@ -85,5 +85,14 @@ describe('Tests for table filter', () => {
     cy.get('[data-cy-debug="clear-filter-btn"]').click()
     cy.get('[data-cy-debug="tableFilter"]').eq(3).type('*Simple*');
     cy.assertDebugTableLength(2);
+  });
+
+  it('Should convert filter to new values when a different view is selected with different columns', () => {
+    cy.get('[data-cy-debug="filter"]').click();
+    cy.get('[data-cy-debug="tableFilter"]').eq(1).type('1{enter}');
```
ricklue, line 92: Value '1' is not a valid EndTime value for this filter right?

MarkvdVorst: It is. As long as it follows part of the pattern of endtime, it should not give an error. Validation for endtime is bit finicky as the use of wildcards was also required.

This means if the date is 13-12-2024 24:30:25 and I put 13-12 or just 13 as the filter it should recognise it as well as part of the date.

ricklue (dismissed):

src/app/debug/filter-side-drawer/filter.service.ts
```diff
@@ -6,17 +6,18 @@ import { debounceTime, filter, Observable, Subject } from 'rxjs';
...
   private metadataTypes: Map<string, string> = new Map<string, string>();
+  private metadataLabels: Array<string> = new Array<string>();
   private filterErrors: Map<string, string> = new Map<string, string>();
 
   showFilter$: Observable<boolean> = this.showFilterSubject.asObservable();
-  metadataLabels$: Observable<string[]> = this.metadataLabelsSubject.asObservable();
+  metadataLabels$: Observable<Array<string>> = this.metadataLabelsSubject.asObservable();
   currentRecords$: Observable<Map<string, Array<string>>> = this.currentRecordsSubject.asObservable();
   metadataTypes$: Observable<Map<string, string>> = this.metadataTypesSubject.asObservable();
   filterError$: Observable<[boolean, Map<string, string>]> = this.filterErrorSubject.asObservable();
```
MatthijsSmets, line 23: Please order these class variables to be in line with [our conventions](https://github.com/frankframework/frontend-conventions/blob/master/typescript.md#class-member-ordering)

src/app/debug/table/table.component.ts
```diff
@@ -205,9 +205,7 @@ export class TableComponent implements OnInit, OnDestroy {
         value = '';
       }
     }
-    this.tableSettings.currentFilters = new Map<string, string>();
-    this.tableSettings.currentFilters = filters;
-
+    this.tableSettings.currentFilters = new Map(filters);
```
MatthijsSmets, line 208: Why create a new `Map` object? 
`filters` already is a `Map`?
```suggestion
    this.tableSettings.currentFilters = filters;
```

MatthijsSmets (changes requested):
Two small things

src/app/debug/table/table.component.ts
MarkvdVorst: This was for debugging and forgot to get rid of it. The value in activeFilters wasn't being updated and this was one of the debugging lines for hard updating it. I'll change it back.

src/app/debug/filter-side-drawer/filter.service.ts
MarkvdVorst: Added the change.

MatthijsSmets (approved):

PR #618, opened 2024-08-19
Removed unused saveReportEvent & closeReportEvent

### LINKED ISSUE #603: saveReportEvent & closeReportEvent never used
In the edit display component, there is a subject called saveReportEvent that is never used. There is only one next call on the subject, but it is never used in a subscribe or in the parent component with method binding.

For the closeReportEvent it is exactly the other way around. It is used with method binding in the report component, but the only next call on the subscribe is dependent of a condition which is always false. See code below:
image
This function is called two times and the argument in this function is always false

ricklue:
Closes #603 
Also removed todos from cypress tests, because the order of the reports in the reports table is now fixed.


src/app/report/report.component.ts
```diff
@@ -77,7 +77,7 @@ export class ReportComponent implements AfterViewInit, OnInit {
     if (this.newTab && this.reportData) {
       this.tabService.closeTab(this.reportData);
     }
-    this.displayComponent.closeReport(false);
+    this.displayComponent.closeReport();
   }
 
   closeReport(): void {
```
MarkvdVorst, line 83: closeReport is not used anymore. Is there still a reason to keep this?

MarkvdVorst (changes requested):
Looks good, just have one question.

MatthijsSmets (approved):

src/app/report/report.component.ts
ricklue: We don't remove one single report from the debug tree, right? If it's no, then there is no reason I think. What do you think @MatthijsSmets?

MatthijsSmets: You can remove this method

MarkvdVorst (approved):

PR #626, opened 2024-08-21
feat: form validation for open reports in progress input element

### LINKED ISSUE #578: It is possible to open report-in-progress indexes that do not exist
I found this issue using ladybug-test-webapp. I created two reports-in-progress. Then I entered index 10 and clicked the button to open the report-in-progress with that index.

image

I got a toast saying that the report-in-progress with index 10 was opened, but there were only two so this can not be correct.

MatthijsSmets:
Input element to open reports in progress is now validated so the user can only enter values for existing reports in progress.
On invalid input the button is disabled.
Closes #578.


ricklue (commented):
image

When I don't open the report in progress yet, but refresh the page first, I get a pop-up that report in progress with index 1 is opened

ricklue (commented):
image
When I open the report in progress and then I close it with the button above the debug tree section, I get a danger toast that the report can't be found, but the report did close in the debug tree. However the editor section is still there.

MatthijsSmets:
Fixed these issues

ricklue (approved):

PR #629, opened 2024-08-23
feat: views in compare tab

### LINKED ISSUE #345: Implement views in Compare tab

### LINKED ISSUE #588: Would be nice to be able to edit in compare
For example in the following it would be nice to add an enter to more easily compare the strings:

image

MatthijsSmets:
Views are now added to compare tab.
Possible views are shown based on the storageName of the reports.                                                                    
This PR contains a small temporary fix that can be removed when [backend issue](https://github.com/wearefrank/ladybug/issues/283) is solved.
Closes #345.
Closes #588.


src/app/test/test.component.ts
```diff
@@ -248,6 +250,35 @@ export class TestComponent implements OnInit {
...
+    //   next: (value) => {
+    //     this.httpService.getReport(report.storageId, this.storageName).subscribe({
+    //       next: (response: Report): void => {
+    //         report = { ...response };
+    //       },
+    //     });
+    //   },
+    //   error: () => catchError(this.errorHandler.handleError()),
+    // });
+  }
```
MarkvdVorst, line 280: these methods was moved to test.table component

MarkvdVorst (changes requested):
Apart from one thing, this looks good

ricklue (commented):
The compare tab starts with default view 'Report name only'. Maybe this must be 'White box'?

src/app/test/test.component.ts
MatthijsSmets: Good catch, fixed it.

MarkvdVorst (approved):

PR #640, opened 2024-08-26
Refactor of test table

### LINKED ISSUE #582: No scrollbars in Test tab
When your window is small or you have a lot to tests you are not able to see / scroll to the bottom tests:

image

### LINKED ISSUE #625: Refactor test tab table
The test tab table should be using the mat table.
Additionally it should not use the bootstrap css classes.

MarkvdVorst:
Closes #582
Closes #625 
Changed the table to a mat-table.
The new table is scrollable.

If table ever needs to be sorted by columns, the mat-sort can be added like with the debug table


cypress/support/commands.ts
```diff
@@ -152,13 +153,13 @@ Cypress.Commands.add('enableShowMultipleInDebugTree' as keyof Chainable, (): voi
...
 Cypress.Commands.add('checkTestTableNumRows' as keyof Chainable, (length: number): void => {
-  cy.get('[data-cy-test="table"] tr').should('have.length', length);
+  cy.getTestTableRows().should('have.length', length);
 });
 
 Cypress.Commands.add('checkTestTableReportsAre' as keyof Chainable, (reportNames: string[]): void => {
   cy.checkTestTableNumRows(reportNames.length);
   for (const reportName of reportNames) {
-    cy.get('[data-cy-test="table"]').find('tr').contains(`/${reportName}`).should('have.length', 1);
+    cy.getTestTableRows().contains(`/${reportName}`).should('have.length', 1);
```
MatthijsSmets, line 162: Add a comment to this command that it won't work with duplicate reports names.

MatthijsSmets:
image
Theres some whitespace under the table, it should fill the rest of the entire page right?

MatthijsSmets:
Also why is the font-size so much bigger compare to the old one?

MarkvdVorst:
> Also why is the font-size so much bigger compared to the old one? Old: image New: image

I lowered the font size, but I'm not exactly sure what caused it to be smaller. Since no font-size was actually defined previously. My guess is that bootstrap had some predefined sizes, but those are not in the table, table-responsive or table-hover classes. Maybe it was in one of the parent container classes that used a bootstrap class.

Also added the comment and gave td elements a max width as without this, the table would take up space where the file tree element is.

MatthijsSmets:
image
The table is not taking the full width

MarkvdVorst:
> image The table is not taking the full width

Changed it so it now takes up the remaining width.

MatthijsSmets (approved):

PR #650, opened 2024-09-04
fix: folder tree in test tab

### LINKED ISSUE #320: Sort folders in Test tab
When reports are divided over several folders in the Test tab the folders shown in the tree on the left and in the content pane on the right should be sorted on alphabet.

Maybe the backend should return the metadata records in the required order?

### LINKED ISSUE #595: Fix folder tree Test tab
As it stands the folder tree only ever shows two folders; the root folder and the last folder tests were moved to.

It should show every folder that is used for the test reports (see picture below)

Image

MatthijsSmets:
Previously folders in the test tab did not update correctly and also just didn't work correctly.
Folder tree will now update as soon as a report is moved.
Folders are now also sorted alphabetically.
Closes #595 
Closes #320

### DIFF
--- a/src/app/test/test-folder-tree/test-folder-tree.component.ts
+++ b/src/app/test/test-folder-tree/test-folder-tree.component.ts
@@ -2,6 +2,7 @@ import { Component, Output, ViewChild } from '@angular/core';
 import { Subject } from 'rxjs';
 import { SimpleFileTreeUtil } from '../../shared/util/simple-file-tree-util';
 import { CreateTreeItem, FileTreeOptions, NgSimpleFileTree, NgSimpleFileTreeModule } from 'ng-simple-file-tree';
+import { TestListItem } from '../../shared/interfaces/test-list-item';
 
 @Component({
   standalone: true,
@@ -13,8 +14,9 @@ import { CreateTreeItem, FileTreeOptions, NgSimpleFileTree, NgSimpleFileTreeModu
 export class TestFolderTreeComponent {
   @ViewChild('tree') tree!: NgSimpleFileTree;
   @Output() changeFolderEvent: Subject<string> = new Subject<string>();
-  rootFolder: CreateTreeItem = {
+  readonly rootFolder: CreateTreeItem = {
     name: 'Reports',
+    path: '',
   };
   treeOptions: FileTreeOptions = {
     folderBehaviourOnClick: 'select',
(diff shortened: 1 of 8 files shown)


src/app/test/test-folder-tree/test-folder-tree.component.ts
```diff
@@ -24,81 +26,59 @@ export class TestFolderTreeComponent {
...
+    this.tree.items = [];
+    this.rootFolder.children = [];
+    for (let item of data) {
+      if (item.path) {
+        this.nextAction(
+          item,
+          item.path.split('/').filter((s) => s !== ''),
+          this.rootFolder,
+        );
+      }
```
Matthbo, line 45: Isn't it better to do this recursively?
It gives the opportunity to actually track the parent and children and make it behave like an actual NodeTree structure, now you're doing it the opposite way making it quite harder for yourself

src/app/test/test-folder-tree/test-folder-tree.component.ts
```diff
@@ -24,81 +26,59 @@ export class TestFolderTreeComponent {
...
   }
 
-  resetTree(): void {
-    if (this.tree) {
-      this.tree.clearItems();
-    }
-    this.originalItems = [];
+  autoSelectItem(item: CreateTreeItem): boolean {
+    return item.path === this.rootFolder.name;
   }
```
Matthbo, line 83: This is only used in `autoSelectCondition` right?
Can't you do it inline in there since you don't use it anywhere else?

Matthbo (changes requested):

src/app/test/test-folder-tree/test-folder-tree.component.ts
MatthijsSmets: How do you mean? `nextAction` is already called recursively

Matthbo: You do use  recursion, but not for the children of the root folder.
Why not use recursion for every folder starting with the root folder itself? Next action already tries to get all children recursively so why resort to also using a for-loop for what is essentially the same logic but closer to the root?

MatthijsSmets: Because it makes folders based on the path of the items, the items are supplied in a 1 dimensional array, so they don't have a parent-child structure yet.

Matthbo: I don't see how that changes anything either way, but you're already splitting the path into route parts right?
`/test/test-a/a` and `/test/test-b/b` would be split up into a root folder with both folders `test-a` and `test-b` and than that recursively deeper right?
What does the for-loop do any different except actually calling the array.split function?

Matthbo: Here is an example of how I would solve it, after talking with Sergi about it
https://www.typescriptlang.org/play/?#code/C4TwDgpgBAlsEFsoF4oG8BQVtQHYEMEIAuKAZ2ACcZcBzAGixzH2AAtSLq7GcoBjNjAA2AE0oRcpOIgDaAXQDcGAL4YM-APa4KUSps3Bp8JKjQEipAOT7DV+lBbtrAensChYiVKgKVyjW1dYRhdVFkmbCsISRdgAHcYl3EYCBcIGlEIe0ioaNiEpIzcLLdeHCt02OKsnL583BcmqsbCtJrs8qiOqwx5dQAzTUooAAotHWBHVjYoTQGoEIoASnRcid0yfFw4GAAvCFEASRMUcm3dg4AFGdGnNmXc-FFjkzIAMX0EACUIYVYYAA3CA3dijLY7YD7Q4nRAAOjIYBCwFGlSsywctmAjzUgR0mmEEDhwk0tFGWMeGAGAFdcPwodpzpDoaC2HcZpwqDRaMtOdxaGs+BJgNTKLhpuw4RIkfh+BBRi4ADouADULgcaOWUogMrlCoAesqAD7KgAk6ry6OUuJpdIZ4uer0QHy+v3+UOBrPJfwBno55C5dAUDhY3iMsBMvKggM0MFEgpwMAW3vdQJBM2Jklo7BQyFQAAZVpg+EKICKxco+Li+BsprgIPEAMKeUTGRBnYsl7AWEh6H0e9PsWT5+QOXJ8e6kABWsdwbt9g7ZockwDh90x-bTrOH8gx45wghE4kkpAUuTUNaCUwgAA9QlC6M2j2dl7hV4evJIAPxwgaZUbjC2qzIAAfB4R5wj2uZ9qmfpDiOjyXpM4FiPOA6smcEiwYuCIhHqACMiGJsmACEt73tyT5iEW+7YK+74tt4a7UmQbL1k2QFQE0AjbFA+DCBIzwgFAoiaOQNByhGZB9gMEDeJJoTTGQZCHLA4rsNAIT1lAABGfyaPEfHSYkwjCLRfblg6LywggLqaD8m5wWyH6iGhW4zA47FUaIRHYBeODCqKVlOrZnz2W5TmAUeEWLg45EUJRQGqIMtL0jAjIzjQMVen+lAUKyfLcg4KkTKIBUBvyNGlpZYy5flGaSKIZAAOpwGympQF+lpQK46JQCq5AQKVrKqEAA

src/app/test/test-folder-tree/test-folder-tree.component.ts
MatthijsSmets: Fixed

src/app/test/test-folder-tree/test-folder-tree.component.ts
MatthijsSmets: Fixed it.

Matthbo (approved):

PR #651, opened 2024-09-05
Allow starting to connect with FF

mhdirkse:


MatthijsSmets:
I think this fits better in the [ladybug-ff-cypress-test](https://github.com/wearefrank/ladybug-ff-cypress-test) repository right?

MatthijsSmets (approved):

PR #658, opened 2024-09-06
fix: fix deleting reports in test tab

MatthijsSmets:
Deleting reports in the test tab now only deletes the one shown in the folder and those that are selected.


Matthbo (dismissed):

Matthbo:
> Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/
I think the github action needs to be updated

Matthbo (approved):

PR #680, opened 2024-09-17
Fix: report alert messages are always shown

### LINKED ISSUE #683: Encoder used to show checkpoint value should always be shown

MatthijsSmets:
Previously alert messages for reports like: `Message is null` or `Message of type ... is encoded using ...` were only shown when clicking the `Show metadata` button.
image

Now they are always shown.
image

Closes #683


src/app/report/report-alert-message/report-alert-message.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  checkIfAnyMessagesPresent(): void {
+    if (ReportUtil.isCheckPoint(this.report)) {
+      this.anyMessagesPresent = !!(
+        this.report.streaming ||
+        this.report.stubbed ||
+        !this.report.message ||
+        this.report.encoding ||
+        (this.report.preTruncatedMessageLength && this.report.preTruncatedMessageLength > 0) ||
+        this.report.stubNotFound
+      );
```
Matthbo, line 31: `!this.report.message` why is this turned negative? Wouldn't not having a message do the exact opposite of what you want?

MatthijsSmets: Yeah the property names are a little confusing, this method sets a boolean to true if any alert messages should be shown. One of those alert messages is that the message of a report is null, thats why it checks if its null. I will change the name of the method.

Matthbo: Maybe also rename `anyMessagesPresent` to `anyAlertMessagesPresent`

MatthijsSmets: Yeah good point, fixed it.

Matthbo (approved):

PR #726, opened 2024-10-15
Fix: show front and backend versions

MatthijsSmets:
Frontend and backend versions are now shown in the settings modal in the debug tab:
image

Frontend version is now also set in the browser tab:
image

Also added unit tests for both the newly created version service, as well as the app component to confirm that title and frontend version are set correctly.


src/app/shared/services/version.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  async getFrontendVersion() {
+    if (!this.frontendVersion) {
+      try {
+        const packageJson = await firstValueFrom(this.httpClient.get<{ version: string }>(this.packageJsonPath));
+        if (packageJson) {
+          this.frontendVersion = packageJson.version;
+        }
+      } catch (error) {
+        console.error('package.json could not be found in assets', error);
```
Matthbo, line 27: If `this.frontendVersion` is undefined it will never in this catch, shouldn't you set it to a default value?
This will break other components that also have set the frontendVersion variable to be able to be undefined but is used within a template

src/app/debug/table/table-settings-modal/table-settings-modal.component.html
```diff
@@ -10,6 +10,20 @@ <h5 class="modal-title" id="settingsModal">Options</h5>
...
+      <div>
+        <h5>Info</h5>
+        <div class="d-flex flex-column gap-2">
+          <div class="d-flex flex-row gap-1">
+            <p class="m-0">Backend Version: {{ backendVersion }}</p>
+            <i class="bi bi-copy cursor-pointer" [appCopyTooltip]="backendVersion"></i>
+          </div>
+          <div class="d-flex flex-row gap-1">
+            <p class="m-0">Frontend Version: {{ frontendVersion }}</p>
+            <i class="bi bi-copy cursor-pointer" [appCopyTooltip]="frontendVersion"></i>
```
Matthbo, line 22: What if any of these values are still undefined because the http request is slow or because the package.json couldn't be found?
This will cause errors that might break rendering the whole component template

Matthbo (changes requested):

src/app/debug/table/table-settings-modal/table-settings-modal.component.html
MatthijsSmets: But Angular handles undefined strings just fine, they are just treated as an empty string.

Matthbo: Okay crazy, that's relatively new because it sure didn't used to be like that
My excuses

Matthbo (approved):

PR #819, opened 2024-12-10
717 error message not shown in debug tab

MatthijsSmets:
Error message is now shown in a toast


Matthbo (approved):

cypress/e2e/no-profile/debug/runningReports.cy.ts
```diff
@@ -59,7 +59,7 @@ describe('Test Reports in progress warning', () => {
 
   it('If threshold time has been met then show warning', () => {
     cy.get('[data-cy-debug="refresh"]').click();
-    cy.contains(`[One or more reports are in progress for more than 1 minutes]`);
+    cy.contains(`[One or more reports are in progress for more than`);
```
Matthbo, line 62: "for more than"...?

MatthijsSmets: Yes this just checks that the error message is there and not the exact value, looking at the cypress videos from the cicd, it shows a value like: 0.00000001166777. This also means that if we change the timeout time in the config for the backend, we won't have to change the cypress tests.

Matthbo: Ah gotcha

Matthbo (approved):

PR #860, opened 2024-12-24
Fix: remove form validation

MatthijsSmets:
Removed form validation on frontend, because of poor implementation.

All validation is now done in the backend. Further refactoring of backend error messages is necessary to show user error feedback.


Matthbo (approved):
Is there still some basic html5 form validation?
Input fields having the right type and not having everything be a text input for example

MatthijsSmets:
No they don't because the backend allows filtering with wildcards and other symbols like <>|

MatthijsSmets:
No, the filters are updated and sent to the backend on `keyup`, so the users gets feedback fairly quickly

Matthbo:
You can't guarantee the response speed of the backend will be very fast in every possible scenario, a user could also spam submit a form or spam click a button which may lead to duplicate requests to the backend. This might be fine when it comes to filtering items as I think is the case here, but disabling inputs until the request is finished is basically just keeping track of 1 boolean worth of effort....

MatthijsSmets:
That's why there's a debounce time on each form field to delay the time before a request to the backend is sent.

Matthbo (approved):

PR #889, opened 2025-02-03
feat: Add option to display message context

alisihab:
#458 

image

### DIFF
--- a/src/app/report/edit-display/edit-display.component.html
+++ b/src/app/report/edit-display/edit-display.component.html
@@ -37,6 +37,13 @@
           (click)="toggleMetadataTable()">
           {{ metadataTableVisible|booleanToString:'Hide' :'Show' }} metadata
         </button>
+        <button
+          data-cy-open-metadata-table
+          class="btn btn-info h-fit"
+          title="Open message context table"
+          (click)="toggleMessageContextTable()">
+          {{ messageContextTableVisible|booleanToString:'Hide' :'Show' }} messagecontext
+        </button>
         <app-toggle-button
           #editToggleButton
           data-cy-report="toggleEdit"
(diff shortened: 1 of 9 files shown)


src/app/shared/interfaces/checkpoint.ts
```diff
@@ -7,6 +7,7 @@ export interface Checkpoint {
   index: number;
   level: number;
   message: string;
+  messageContext: Map<string, string>;
```
MatthijsSmets, line 10: I'd replace the map with a Record<string, string> [see why](https://stackoverflow.com/questions/56398223/differences-between-and-when-to-use-map-vs-record)

src/app/report/edit-display/edit-display.component.html
```diff
@@ -37,6 +37,13 @@
           (click)="toggleMetadataTable()">
           {{ metadataTableVisible|booleanToString:'Hide' :'Show' }} metadata
         </button>
+        <button
+          data-cy-open-metadata-table
+          class="btn btn-info h-fit"
+          title="Open message context table"
+          (click)="toggleMessageContextTable()">
+          {{ messageContextTableVisible|booleanToString:'Hide' :'Show' }} messagecontext
+        </button>
```
MatthijsSmets, line 46: Maybe hide the button if the selected node is not a checkpoint?

MatthijsSmets (commented):
Looks good, just some small remarks. Also have a look at the eslint warnings/errors.

MatthijsSmets (dismissed):

src/app/shared/components/messagecontext-table/messagecontext-table.component.html
```diff
@@ -0,0 +1,13 @@
+<div class="d-flex flex-column mb-2">
+  @if (messageContextData.length > 0) {
+    <div class="title">MessageContext:</div>
+    <div data-cy-metadata-table="table" class="metadata-table">
+        @for (entry of messageContextData; track entry) {
+        <div class="table-col first-table-col">
+          <div class="key">{{ entry[0] }}</div>
```
Matthbo, line 7: Isn't the key unique so it can be used to track in the for-loop?

Matthbo (approved):

Issue #918, opened 2025-02-25
Next Ladybug release in the FF!


mhdirkse:
I propose to keep reminders here. Reminders for what we postpone until a new release of Ladybug in the FF!. Let us add comments for anything we want to postpone until that moment.

mhdirkse:
The data-cy* labels are not as should be, see picture below:

![Image](https://github.com/user-attachments/assets/71a99a71-d929-407f-af9a-c7bdac72af6c)

Both have `data-cy-open-metadata-table`. The button marked red should have another data-cy* tag.

PR #920, opened 2025-02-26
Make messages context table custom attribute unique

alisihab:


mhdirkse:
Merging should be postponed until we do a release of Ladybug in the FF!. See https://github.com/wearefrank/ladybug-frontend/issues/918.

mhdirkse (approved):

PR #934, opened 2025-03-13
Use ESLint 9

mhdirkse:


mhdirkse:
I temporarily added dependency `@eslint/migrate-config` in package.json. Probably I then did `npx  @eslint/migrate-config .eslintrc.json` to get started with the migration. I debugged the result until the prettier rules and the unicorn rules worked. I checked both in CI/CD and locally by temporarily introducing linting errors.

philipsens:
I am currently working on recreating the config in the FF! Flow. You could maybe use that one when it's done.

mhdirkse:
Thanks. I will wait.

philipsens:
Hi,

I've created a new EsLint Config for in the FF! Flow. You can find it here: https://github.com/frankframework/flow/blob/master/src/main/frontend/eslint.config.mjs

I think a lot of the rules are the same, you can add anything that is missing.

mhdirkse:
Superseded by https://github.com/wearefrank/ladybug-frontend/pull/943.

PR #936, opened 2025-03-13
Update Angular to 19.2.2

mhdirkse:


mhdirkse:
This is too quick. I put versions in package.json that do not exist. Better to approve patches from the Dependabot PRs.

PR #1081, opened 2025-08-26
To angular 20

mhdirkse:


Matthbo (approved):

Matthbo:
Looks like `ng-simple-file-tree` and `ngx-monaco-editor-v2` need to be updated or tested if they are able to support Angular 20.2

mhdirkse:
I found an issue in the PR as it is at this point. The text of a checkpoint as become editable. Editen van een rapport gaat ook niet meer.

PR #1082, opened 2025-08-28
To angular 20 and update monaco

mhdirkse:


mhdirkse:
We have to get rid of ngx-monaco-editor-v2

PR #1105, opened 2025-09-29
Bump Monaco editor to 0.55.1

mhdirkse:


mhdirkse:
When we update to 0.55.1 then there is no build problem like is the case with 0.53.0. There fails a test because checkpoint values shown with the Monaco editor are not editable.

mhdirkse:
At this point, the ladybug-frontend tests succeed but not ladybug-ff-cypress-test. The jobs involving the test webapp fail. These tests use a custom pipe to create a report in a before() handler. The tests then fail because no ladybug report is found. When I look in the log the adapter call succeeds, but it does not produce the report.

